PR #24, opened 2014-12-24
Add loop

anvanza:
There was a comment that said ('Here if somebody could find how to do it easily for all the flags it would be awesome') , so I did.


rande:
@nicolas-bastien ^ ;)

nicolas-bastien:
@anvanza thanks !

PR #32, opened 2015-03-16
Fix new corebundle errorElement

huitiemesens:


nicolas-bastien:
I'll check that after the other PR.

I'll try to find a way to handle that without BC break...

rande:
@nicolas-bastien you can remove the empty methods ;)

nicolas-bastien:
@huitiemesens as @rande said you can simply remove this empty method validateBlock and buildEditForm so it will works with no BC breaks.

@rande lol I was not uptodate with this commit : https://github.com/sonata-project/SonataBlockBundle/commit/c3488cd9f5686d2c9eae105608d384b9afbb1866
thanks ;)

nicolas-bastien:
@huitiemesens why did you close this ?

I just wait for test result and merge

huitiemesens:
@nicolas-bastien as rande said , i though you had already fix this without this PR

nicolas-bastien:
@huitiemesens I could, but as you began this I think I think it's fair to let you finish.

I was just joking with him about its so simple solution I didn't noticed ;)

Thanks

By the way did you noticed I made a release 1.0.0 so you should change the versiong on your project if you want to be uptodate...

huitiemesens:
@nicolas-bastien  initially i forked your project to fit a few thing according to a project which has been canceled , so i think i ll closed this repo , because it s not needed anymore :) 

but i ll continue to help in sonata-project as much as i can, i really like your projects

PR #33, opened 2015-03-16
Prepare 1.0.0 release

nicolas-bastien:


rande:
@nicolas-bastien I think the 1.0.0 should use the old ErrorEelement

rande:
the 2.0.0 should use the new version ;)

nicolas-bastien:
Hi @rande,

We spoke about it a long time ago.
I'd like to tag before merging the next PR, stil ok with this versionning ?

nicolas-bastien:
ok you're too fast !

PR #44, opened 2015-06-18
[Fix] csslint standard rules disallow "0px" syntax on css files.

nicolas-joubert:
Hello,
i'm using your assets with grunt-csslint then standard rules (and css good practices) says that 0 values must not have unit.
NB : css was been redo with grunt-less (lessc inside)

### DIFF
--- a/Resources/public/css/sonata-translation.css
+++ b/Resources/public/css/sonata-translation.css
@@ -1,25 +1,22 @@
 /**
  * SonataTranslation: adds translations to your forms
  */
-.sonata-bc {
-  /** Here if somebody could find how to do it easily for all the flags it would be awesome */
-}
 .sonata-bc .sonata-ba-form .locale_switcher {
-  padding: 1px 0px 0px 0px;
-  text-align: right;
-  margin-right: 15px;
-  float: right;
+    padding: 1px 0 0 0;
+    text-align: right;
+    margin-right: 15px;
+    float: right;
 }
 .sonata-bc .sonata-ba-form .locale_switcher a {
-  margin-right: 0px;
-  padding: 2px;
-  opacity: 0.5;
+    margin-right: 0;
+    padding: 2px;
+    opacity: 0.5;
 }
 .sonata-bc .sonata-ba-form .locale_switcher a.active {
-  opacity: 1;
+    opacity: 1;
 }
 .sonata-bc .sonata-ba-form .locale_switcher a:hover {
-  opacity: 1;
+    opacity: 1;
 }
 .sonata-bc .sonata-ba-form label.wysiwyg.locale,
 .sonata-bc .sonata-ba-form label.checkbox.locale,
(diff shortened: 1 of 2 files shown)


nicolas-bastien:
@nicolas-joubert whould be so much better if you reindent all in a separate commit

so every body can see clearly your modification in history ....

waiting on travis

Soullivaneuh:
@nicolas-joubert 4 spaces indent? From which rules is coming from? Don't really like this approach.

Otherwise, the idea is nice. Would be better if it could be automated like I did here: https://github.com/sonata-project/SonataAdminBundle/pull/3038

:+1:

nicolas-joubert:
I've redo CSS with 2 spaces indentation & remove empty rule (which is another css good practice)

Soullivaneuh:
@nicolas-joubert please squash your commits.

What about the automation idea? Could be possible?

@nicolas-bastien I think we should call @rande opinion for this, isn't it?

nicolas-bastien:
@nicolas-joubert I didn't say that 4 spaces is a bad idea, make sense if we have a look at SonataAdmin styles : https://github.com/sonata-project/SonataAdminBundle/blob/master/Resources/public/css/layout.css

I writed that it's bad to do indent + fix in the same commit. 
Code review is a pain and it's not easy for people who want to understand your fix in history.

@Soullivaneuh I'm goin in holiday for 10 days and don't have time to look at automation.
See that with rande directly if it's something you're working on together or we'll see that when I'm back

Soullivaneuh:
> I didn't say that 4 spaces is a bad idea, make sense if we have a look at SonataAdmin styles.

Didn't say that neither. :-)

But for sure, we have to choose a coding standard for that.

I'll up-vote for 2 spaces because of many front-end project is following this rules but it's not a locked discussion.

> I'm goin in holiday for 10 days and don't have time to look at automation.
See that with rande directly if it's something you're working on together or we'll see that when I'm back

Of course. Have to finish PHP CS review on all projects but it can be the next goal.

I'll talk about it with @rande. Have nice holidays! ;-)

nicolas-joubert:
I'm looking for automate csslint on travis today.

nicolas-joubert:
Adding csslint to travis is done there : https://github.com/nicolas-joubert/SonataTranslationBundle/pull/2
...but not compatible with container-based mode and not required to the initial PR which is to correct less & css files.
What's your opinion ?

nicolas-joubert:
It seems to be not possible to add csslint on travis build because enabling sudo will break cache system ; so this PR is finished and ready to Merge if all are OK.

Soullivaneuh:
> It seems to be not possible to add csslint on travis build because enabling sudo will break cache system ; 

Or we could simply add it without sudo using. node and npm can be easily installed on local.

I will keep it on mind to work on it, thanks for the effort. :+1:

PR #46, opened 2015-08-17
Use ManagerType from Admin configuration

wodka:


rande:
ok to test

nicolas-bastien:
@wodka this is not working, manager type is set in the compiler pass with the value declared when you register your admin.

So manager type is "orm" and not "doctrine".

Your code does not work cause orm is not a service.

Can you give me a working solution or can we close this ?

wodka:
@nicolas-bastien oh, it seems I only used doctrine_mongodb

```(php)
/**
 * @DI\Service("admin.cart")
 * @Sonata\Admin(
 *   class="Model\Cart",
 *   managerType="doctrine_mongodb",
 *   group="Order",
 *   label="Carts"
 * )
 */
```


I will now add a check if it is set to orm to use "doctrine" then, this will give support for both odm and orm :)

nicolas-bastien:
ok @wodka thanks !

Nice to see that your working with mongo, would be great if you can add a new page on the doc to explain how to do this.

Like the orm page.

Can you do that in this PR please ?

Admin/Extension/Gedmo/TranslatableAdminExtension.php
```diff
@@ -40,6 +40,22 @@ protected function getTranslatableListener(AdminInterface $admin)
 
         return $this->translatableListener;
     }
+    
+    /**
+     * @param AdminInterface $admin
+     * 
+     * @return \Doctrine\Common\Persistence\ObjectManager
```
nicolas-bastien: can you add a use and just write : @return ObjectManager

-> please append your commits and push force to your repo so history is cleaner

core23:
Ping @wodka

OskarStark:
@wodka please rebase and @nicolas-bastien can you give us a final review of this PR please?
thank you

wodka:
sry, I don't think this is working anymore - I tried now for a few hours but I cannot get it to work

PR #47, opened 2015-08-18
More semver compliant composer configuration

Soullivaneuh:


rande:
@Soullivaneuh CS are broken

Soullivaneuh:
@rande I don't know what you are talking about. ;-)

PR #56, opened 2015-10-08
Update TranslationBundle doc with the right bundle name

cuginolancio:
Fix bundle name


OskarStark:
Thank you @cuginolancio

Issue #55, opened 2015-10-02
Edit Forms not working after upgrade to 1.0.1


Mangetsu:
I'm getting this error after upgrading to Sonata Translation 1.0.1 :
```
CRITICAL - Uncaught PHP Exception Twig_Error_Runtime: "An exception has been thrown during the rendering of a template ("Unable to find template "".") in "SonataAdminBundle:CRUD:base_edit_form.html.twig"."
```

here is my Composer.json :

```
"php":                                        ">=5.3.9",
    "symfony/symfony":                            "2.6.*",
    "doctrine/orm":                               "~2.2,>=2.2.3,<2.5",
    "doctrine/dbal":                              "<2.5",
    "doctrine/doctrine-bundle":                   "~1.4",
    "symfony/assetic-bundle":                     "~2.3",
    "symfony/swiftmailer-bundle":                 "~2.3",
    "symfony/monolog-bundle":                     "~2.4",
    "sensio/distribution-bundle":                 "~4.0",
    "sensio/framework-extra-bundle":              "~3.0,>=3.0.2",
    "incenteev/composer-parameter-handler":       "~2.0",

    "sonata-project/core-bundle":                 "^2.3",
    "sonata-project/admin-bundle":                ">=2.3.3",
    "sonata-project/user-bundle": 				  "~2.2",
    "sonata-project/intl-bundle":                 "^2.2",
    "sonata-project/media-bundle":                "~2.2",
    "sonata-project/doctrine-orm-admin-bundle":   ">=2.3.3",
    "sonata-project/translation-bundle":          "<=1.0.0",
    "sonata-project/formatter-bundle":            "^2.3",
    "pixassociates/sortable-behavior-bundle":     "0.1.*@dev",

    "stof/doctrine-extensions-bundle":            "^1.1",
    "besimple/i18n-routing-bundle":               "^2.3",

    "friendsofsymfony/user-bundle":               "~1.3",

    "doctrine/data-fixtures":                     "~1.0",
    "doctrine/doctrine-cache-bundle":             "~1.0",
    "doctrine/migrations":                        "dev-master",
    "doctrine/doctrine-migrations-bundle":        "dev-master",
    "sonata-project/cache-bundle":                "~2.1",

    "eko/feedbundle":                             "~1.0",
    "lunetics/locale-bundle":                     "~2.3",
    "liip/imagine-bundle":                        "~1.0",
    "wjzijderveld/check-bundles":                 "~1.1",
[...]

PR #57, opened 2015-10-08
Workaround to fix issue #55 introduced in version 1.0.1

cuginolancio:
Add deprecated method to fix bug introduced with commit 8bbcc4e

### DIFF
--- a/Block/LocaleSwitcherBlockService.php
+++ b/Block/LocaleSwitcherBlockService.php
@@ -15,12 +15,21 @@ use Sonata\BlockBundle\Block\BaseBlockService;
 use Sonata\BlockBundle\Block\BlockContextInterface;
 use Symfony\Component\HttpFoundation\Response;
 use Symfony\Component\OptionsResolver\OptionsResolver;
+use Symfony\Component\OptionsResolver\OptionsResolverInterface;
 
 /**
  * @author Nicolas Bastien <nbastien.pro@gmail.com>
  */
 class LocaleSwitcherBlockService extends BaseBlockService
 {
+    /**
+     * {@inheritdoc}
+     */
+    public function setDefaultSettings(OptionsResolverInterface $resolver)
+    {
+        $resolver->configureSettings($resolver);
+    }
+
     /**
      * {@inheritdoc}
      */


Soullivaneuh:
@cuginolancio Which bug? This should work as this.

Do you have an issue reference?

OskarStark:
which symfony version are you using @cuginolancio 

``composer show -i | grep symfony``

and 

``composer show -i | grep sonata``

cuginolancio:
@Soullivaneuh I had the same issue  #55 where the `LocaleSwitcherBlockService::configureSettings()` isn't called:

```
An exception has been thrown during the rendering of a template ("Unable to find template "".") in "SonataAdminBundle:CRUD:base_edit_form.html.twig".
```


@OskarStark those are my dependencies

Symfony: 
```
symfony-cmf/block-bundle                   1.2.2             Symfony CMF Block Bundle
symfony-cmf/content-bundle                 1.2.0             Symfony CMF Content Bundle
symfony-cmf/core-bundle                    1.2.0             Symfony CMF Core Bundle
symfony-cmf/create-bundle                  1.2.2             Symfony Bundle for createphp and create.js. The easiest way to make any site editable and have semantic annotations with RDFa.
symfony-cmf/media-bundle                   1.2.x-dev 783fa6b Symfony CMF Media Bundle
symfony-cmf/menu-bundle                    1.2.0             Symfony CMF Menu Bundle
symfony-cmf/routing                        1.3.0             Extends the Symfony2 routing component for dynamic routes and chaining several routers
symfony-cmf/routing-auto                   1.0.1             Component for automatically creating and managing routes for persisted objects
symfony-cmf/routing-auto-bundle            1.0.0             Bundle which automatically creates and manages routes for persisted objects
symfony-cmf/routing-bundle                 1.3.5             Symfony RoutingBundle
symfony-cmf/seo-bundle                     1.1.2             Symfony CMF Search Engine Optimization Bundle
symfony-cmf/simple-cms-bundle              1.2.0             A simple CMS bundle based on the Symfony CMF
symfony-cmf/symfony-cmf                    1.2.0             Symfony Content Management Framework
symfony-cmf/tree-browser-bundle            1.1.0             Symfony CMF Tree Browser Bundle
symfony/assetic-bundle                     v2.7.0            Integrates Assetic into Symfony2
symfony/monolog-bundle                     2.8.1             Symfony MonologBundle
symfony/swiftmailer-bundle                 v2.3.8            Symfony SwiftmailerBundle
symfony/symfony                            v2.7.5            The Symfony PHP framework
```

Sonata:
```
sonata-project/admin-bundle                2.3.6             Symfony SonataAdminBundle
sonata-project/block-bundle                2.2.13            Symfony SonataBlockBundle
sonata-project/cache                       1.0.6             Cache library
sonata-project/core-bundle                 2.3.9             Symfony SonataCoreBundle
sonata-project/doctrine-extensions         1.0.2             Doctrine2 behavioral extensions
sonata-project/doctrine-orm-admin-bundle   2.3.3             Symfony Sonata / Integrate Doctrine ORM into the SonataAdminBundle
sonata-project/doctrine-phpcr-admin-bundle 1.2.3             Symfony Sonata / Integrate Doctrine PHPCR into the SonataAdminBundle
sonata-project/easy-extends-bundle         2.1.9             Symfony SonataEasyExtendsBundle
sonata-project/exporter                    1.4.0             Lightweight Exporter library
sonata-project/formatter-bundle            2.3.3             Symfony SonataFormatterBundle
sonata-project/google-authenticator        1.0.2             Library to integrate Google Authenticator into a PHP project
sonata-project/intl-bundle                 2.2.3             Symfony SonataIntlBundle
sonata-project/jquery-bundle               1.8.3             Symfony SonatajQueryBundle
sonata-project/media-bundle                2.3.3             Symfony SonataMediaBundle
sonata-project/notification-bundle         2.3.0             Symfony SonataNotificationBundle
sonata-project/seo-bundle                  1.1.10            Symfony SonataSeoBundle
sonata-project/translation-bundle          1.0.1             SonataTranslationBundle
sonata-project/user-bundle                 2.2.4             Symfony SonataUserBundle
```

Soullivaneuh:
Your `sonata-project/block-bundle` is wrong, please upgrade to 2.3.

Version constraint update is needed on `composer.json` indeed. This workaround can't be merged.

cuginolancio:
Ok but I can't upgrade to 2.3 because I'm using the symfony cmf that on `symfony-cmf/block-bundle` requires `sonata-project/block-bundle: >=2.2.8,<=2.2.13`

Soullivaneuh:
That is a really strange version constraint.

ping @rande

Skuzy:
sonata-project/admin-bundle 2.3.6 requires sonata-project/block-bundle ~2.2,>=2.2.7,<2.3 -> no matching package found.

Th3Mouk:
I've got the same problem while upgrading 1.0.1.

My block bundle constraint is due to the page/bundle (2.3.9) : sonata-project/block-bundle: ~2.2,<2.3

OskarStark:
ping @sagikazarmark do you have this problem, too?

@core23 any idea what happend here?

@cuginolancio a test case would be awesome, but normally we fixed this here:
https://github.com/sonata-project/SonataBlockBundle/blob/01c7230db5170d6473c1f8fadb565e454387c9e5/Block/AbstractBlockService.php#L27

seems to be a strange version problem, but i think we can merge this PR to fix this Problem, don't we?@sonata-project/contributors what do you think?

sagikazarmark:
I gave up using the latest stable long time ago, I sticked to development version.

Actually the large number of repositories and the large number of dependencies make this unavoidable.

Not sure if this worth the hassle. I would rather make people using really old versions monkey patch theirs.

I am a big fan of BC, but in this case, it is nearly an impossible thing.

core23:
IMO we should accept this change and remove it later on the next major release, because the BlockManager dependens on the old ``setDefaultSettings`` defintion.

I'm with @sagikazarmark that there are too many (confusing) versions with strange dependencies. I hope that you're release plan will solve the most annoying problems with the next 2.4 releases.

Th3Mouk:
IMO patch release first of all must be BC and devoided of new majors features.

Projects in production must be based on stable release, so this tag shall in no circumstances block.
Develpoment last for months, the 2.4 that will solve major of compatibitily problems, and the versionning concept.

IMO since 2.3 we could have come out 2.4, 2.5, etc.. And actually it's more a 3.0 on master that a 2.4.

sagikazarmark:
> IMO we should accept this change and remove it later on the next major release

I disagree. This change is not in the latest stable release (or is it?), it is in a development version. Also, this is a deprecated method, which has already been removed from all other repositories. Adding it back here is a bad idea IMO. Not to mention that the problem is caused by some weird dependency hell.

I am reluctant to accept this as a solution for the development version. It could be however cherry picked and added to the last stable (2.2.x?). Still, the problem is caused by the dependency, so adding back deprecated code because of this is not a real solution.

nicolas-bastien:
@cuginolancio first sorry to be so long to checkout your problem.

I agree with  @sagikazarmark this is a patch to work with older code so I'm don't want to merge this.

I have to admit that versionning is hard to maintain so as this bundle is pretty simple I think the best way to move on is to fit to last sonata version and let people using older version to override this service.

You can do that easily with a compiler pass, look at http://symfony.com/doc/current/cookbook/bundles/override.html

-> OverrideServiceCompilerPass
and override LocaleSwitcherBlockService

Note : don't use parameters as we gonna remove it to fit sf best practice.

nicolas-bastien:
see #55

PR #59, opened 2015-10-25
Also check for PersonalTranslatable

jorrit:
`Sonata\TranslationBundle\Traits\Gedmo\PersonalTranslatable` uses `Sonata\TranslationBundle\Traits\Translatable`, but `class_uses` will only return `PersonalTranslatable`, although it indirectly uses `Translatable`. I suggest also checking for `PersonalTranslatable` in `TranslatableChecker`. Maybe the traits could even be made configurable like the interfaces or the models.

### DIFF
--- a/Checker/TranslatableChecker.php
+++ b/Checker/TranslatableChecker.php
@@ -72,7 +72,11 @@ class TranslatableChecker
         }
 
         if (function_exists('class_uses')) {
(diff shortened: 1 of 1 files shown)


nicolas-bastien:
Hi @jorrit thanks you're right I merge this.

Not sure it's a good idea to make traits configurable.
I really think working with interface is a better idea.

Thanks for your help

PR #78, opened 2016-03-04
[PHPCR] Make it possible to create new languages of a document

WouterJ:
Currently, it is impossible to add a new translation of a document using the admin fields. With this patch, adding new translations is possible.

Fixes https://github.com/symfony-cmf/cmf-sandbox/issues/323

### DIFF
--- a/Admin/Extension/Phpcr/TranslatableAdminExtension.php
+++ b/Admin/Extension/Phpcr/TranslatableAdminExtension.php
@@ -40,8 +40,21 @@ class TranslatableAdminExtension extends AbstractTranslatableAdminExtension
         $unitOfWork         = $documentManager->getUnitOfWork();
 
         if ($this->getTranslatableChecker()->isTranslatable($object) && ($unitOfWork->getCurrentLocale($object) != $locale)) {
-            //@dbu : doesn't work - $documentManager->bindTranslation($object, $locale);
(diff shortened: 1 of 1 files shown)


maity-kgz:
@WouterJ This patch fixes the issue partially. I.e its actually create new translation and saves to db.
But user interface still is not clear. When you trying to add translation to existing document it is hard to determine what language you are adding. (you can determine current language only by looking to URL in address bar). Highlighted flag is wrong, it is fallback to original language flag.
Issue affects only existing documents.
Creating new document is OK, flags highlighted correctly until you save and try to add other language.

WouterJ:
Thanks for the testing @maity-kgz. I've now changed the code to simply `findTranslation()` + `bindTranslation()` (to set the locale field). It seems to work correctly on my side.

@dbu can you maybe quickly validate this? There was a comment saying this didn't work, but I think they used `bindTranslation` only (without `findTranslation`).

maity-kgz:
@WouterJ  :+1:

WouterJ:
Is anyone maintaining this bundle and willing to merge this fix and create a new release?

dbu:
i am not exactly a maintainer of this, but allowed to merge and the PHPCR part indeed works like this. thanks a lot wouter!

PR #80, opened 2016-03-15
Update composer.json

jdumaz:


composer.json
```diff
@@ -30,7 +30,7 @@
         "psr-4": { "Sonata\\TranslationBundle\\": "" }
     },
     "suggest": {
-        "stof/doctrine-extensions-bundle": "~1.1",
+        "stof/doctrine-extensions-bundle": "~1.1 if you want to use ",
```
greg0ire: If you want to use what?

greg0ire:
> Update composer.json 

This isn't a very good message. You don't write `update file.php` each time you change a file, do you ?

Here are some good recommendations : http://rakeroutes.com/blog/deliberate-git/

PR #82, opened 2016-03-29
make gedmo extension changes backwards compatible

dbu:
there has been a BC break after 1.0.1 that has not been released yet.

### DIFF
--- a/Admin/Extension/Gedmo/TranslatableAdminExtension.php
+++ b/Admin/Extension/Gedmo/TranslatableAdminExtension.php
@@ -26,19 +26,25 @@ class TranslatableAdminExtension extends AbstractTranslatableAdminExtension
      */
     protected $translatableListener;
 
-    public function __construct(TranslatableChecker $translatableChecker, TranslatableListener $translatableListener)
+    public function __construct(TranslatableChecker $translatableChecker, TranslatableListener $translatableListener = null)
     {
         parent::__construct($translatableChecker);
         $this->translatableListener = $translatableListener;
     }
 
     /**
-     * @param AdminInterface $admin
+     * @param AdminInterface $admin Deprecated, set TranslatableListener in the constructor instead.
      *
      * @return TranslatableListener
      */
-    protected function getTranslatableListener()
+    protected function getTranslatableListener(AdminInterface $admin)
     {
+        if (null === $this->translatableListener) {
+            $this->translatableListener = $this->getContainer($admin)->get(
+                'stof_doctrine_extensions.listener.translatable'
+            );
+        }
+
         return $this->translatableListener;
     }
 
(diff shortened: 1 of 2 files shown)


Admin/Extension/Gedmo/TranslatableAdminExtension.php
```diff
@@ -26,19 +26,25 @@ class TranslatableAdminExtension extends AbstractTranslatableAdminExtension
...
     }
 
     /**
-     * @param AdminInterface $admin
+     * @param AdminInterface $admin Deprecated, set TranslatableListener in the constructor instead.
      *
      * @return TranslatableListener
      */
-    protected function getTranslatableListener()
+    protected function getTranslatableListener(AdminInterface $admin)
```
dbu: re-adding this. this restores BC with 1.0.1 which was tagged on september 29. if anybody used the dev version since then and is overwriting this method, this fix is in turn a BC break for them. but i think that is ok - they where using a dev version so its to be expected.

greg0ire: dev versions can indeed break anything, since we cannot tag them

greg0ire: Shouldn't `$admin` be nullable? I think I saw a commit with `$admin = null`, what happened to it?

dbu: yeah i overwrote it after looking at the diff. if we want to be strict about BC, it needs to be like this. otherwise if somebody overwrote getTranslatableListener with the parameter, the signature will not be compatible anymore

greg0ire: I don't think it actually is **incompatible**. I'll test and get back to you.

greg0ire: I think I'm right, the following code raises no error:

```php
<?php

class A
{
    public function test($param = null)
    {
    }
}

class B extends A
{
    public function test($param, $secondParam)
    {
    }
}

$b = new B();
```

dbu: only because those are untyped parameters which allow null implicitly.

https://3v4l.org/CXH5G

Soullivaneuh: @dbu so you mean `AdminInterface $admin = null` wont permit to declare child method without the parameter?

greg0ire: Oh indeed, I did not understand well how it works then. Didn't know the LSP was enforced at the php level.

dbu: @Soullivaneuh not sure. but it for sure does not allow the child method with `AdminInterface $admin` (not nullable). which is part of the BC break we are trying to fix here...

greg0ire: But I get it now, if we call A with `null`, we should be able to call and child class of it with `null` and only that, it makes sense.

Issue #81, opened 2016-03-29
Translation on list


mcrinquand:
Hello,

You recently added the translation for the object show. 

I need the same on the object list. 

I also need the filters to use the translation on list, filter on every locale available...

Thanks.

nicolas-bastien:
@mcrinquand as we said this morning, implement it on your project and share the result here.

Then we will integrate it in this bundle directly.

Issue #61, opened 2015-11-03
TranslationBundle overrides settings from stofBundle


liquorvicar:
In \Sonata\TranslationBundle\Admin\Extension\Gedmo\TranslatableAdminExtension, we are retrieving the TranslatableListener from the stofBundle and changing some settings on it. This caused us an issue on our project as we wanted to be able to set the TranslatableLocale and the Fallback boolean separately from the values that are hardcoded here. We achieved this by overridding the "sonata_translation.admin.extension.gedmo_translatable.class" parameter with the name of our own class.

I'm wondering if perhaps this should be tidied up. Maybe we could either just let the stof Listener do its thing and allow it be configured separately or possibly introduce our own config settings in this bundle which themselves could inherit the stof settings. I think the former would be my preference (as that's how we dealt with the issue in our project).

What do people think? I'm happy to start a PR if you think this would a valuable change. This would also resolve the issue in https://github.com/sonata-project/SonataTranslationBundle/issues/58 as we wouldn't need to load the private service from stof.

dbu:
i am not promising to take over maintainance of this bundle, but i am currently looking over the open issues and try to do a bit of cleanup.

the code in the gedmo admin extension seems to fix gedmo to the requested locale (which might be different from the request locale, e.g. an english person looking at a french article does not want the navigation to switch to french as well...). i am not sure we can get rid of this. 

if the admin extension has some post save callback, we might be able to safe the things from the translatable listener and restore them after saving... or am i misunderstanding things? another way out could be to make language pinning optional - but i am not sure of the side effects.

OskarStark:
@dbu this issue is very old and still no feedback from @liquorvicar, we should close this issue

PR #83, opened 2016-03-29
#81 add translation on list and on filters

mcrinquand:
In the configureQuery, with the translatable listener the locale is changed for Doctrine and data are translated on the list. 

translationFieldFilter is a callback method to use on doctrine_orm_callback filter to filter on field and on its translation.

### DIFF
--- a/Admin/Extension/Gedmo/TranslatableAdminExtension.php
+++ b/Admin/Extension/Gedmo/TranslatableAdminExtension.php
@@ -11,10 +11,10 @@
 
 namespace Sonata\TranslationBundle\Admin\Extension\Gedmo;
 
-use Doctrine\DBAL\Query\QueryBuilder;
 use Gedmo\Translatable\TranslatableListener;
 use Sonata\AdminBundle\Admin\AdminInterface;
 use Sonata\AdminBundle\Datagrid\ProxyQueryInterface;
+use Sonata\DoctrineORMAdminBundle\Datagrid\ProxyQuery;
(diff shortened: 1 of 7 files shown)


Admin/Extension/Gedmo/TranslatableAdminExtension.php
```diff
@@ -55,4 +57,61 @@ public function alterObject(AdminInterface $admin, $object)
...
         }
     }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function configureQuery(AdminInterface $admin, ProxyQueryInterface $query, $context = 'list')
+    {
+        $this->getTranslatableListener()->setTranslatableLocale($this->getTranslatableLocale($admin));
+        $this->getTranslatableListener()->setTranslationFallback('');
```
dbu: @liquorvicar i guess this would add more to the problem you report in #61 - then again i think the fix likely will look the same as for the other occurance of this code.

dbu:
looks good to me. i am not familiar with the sonataci quick check - that one seems to fail.

greg0ire:
Run `make cs` to fix the build.

greg0ire:
> i am not familiar with the sonataci quick check

@dbu : you have to click on "Console Output" to see something useful

dbu:
continued in #84

dbu:
oh, and @mcrinquand: please comment something after updating a pull request. i only get email notifications on comments, not on new commits, so there is quite a risk i would not see new commits.

PR #91, opened 2016-04-16
Fixed necessary dependencies

greg0ire:
add missing sonata admin bundle dependency and require only the needed
symfony components.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -25,16 +25,19 @@
         "twig/twig": "~1.23"
     },
     "require-dev": {
-        "stof/doctrine-extensions-bundle": "~1.1",
         "knplabs/doctrine-behaviors": "~1.0",
+        "stof/doctrine-extensions-bundle": "~1.1",
         "symfony/phpunit-bridge": "~2.7|~3.0"
     },
     "autoload": {
         "psr-4": { "Sonata\\TranslationBundle\\": "" }
     },
+    "config": {
+        "sort-packages": true
+    },
     "suggest": {
-        "stof/doctrine-extensions-bundle": "~1.1",
         "knplabs/doctrine-behaviors": "~1.0",
+        "stof/doctrine-extensions-bundle": "~1.1",
         "sonata-project/doctrine-orm-admin-bundle":"~2.2",
         "sonata-project/doctrine-phpcr-admin-bundle": "~1.0"
     },
--- a/composer.json
+++ b/composer.json
@@ -18,8 +18,11 @@
     ],
     "require": {
         "php": ">=5.4.0",
-        "symfony/symfony": "~2.2|~3.0",
-        "sonata-project/core-bundle": "^2.2.1"
+        "sonata-project/admin-bundle": "^2.3.0|~2.4@dev",
+        "sonata-project/core-bundle": "^2.2.1",
+        "symfony/framework-bundle": "~2.2|~3.0",
+        "symfony/options-resolver": "~2.2|~3.0",
+        "twig/twig": "~1.23"
     },
     "require-dev": {
         "stof/doctrine-extensions-bundle": "~1.1",


composer.json
```diff
@@ -18,9 +18,14 @@
     ],
     "require": {
         "php": ">=5.4.0",
-        "symfony/symfony": "~2.2|~3.0",
+        "symfony/framework-bundle": "~2.2|~3.0",
+        "symfony/options-resolver": "~2.2|~3.0",
+        "symfony/twig-bridge": "~2.2|~3.0",
+        "sonata-project/admin-bundle": "~2.3",
```
core23: Can we make this optional? 

IMO the bundle would work without the ``SonataAdminBundle``, but we have to move the ``sonata.timeline.admin.extension`` to a seperate file.

greg0ire:
Wow. This fails because of a timeout. I will retry it.

greg0ire:
When I try that locally, I don't have enough memory…

.travis.yml
```diff
@@ -24,7 +24,7 @@ matrix:
       env: TARGET=cs_dry_run
     - php: 7.0
       env: TARGET=docs
-    - php: 5.4
+    - php: 7.0
```
core23: This won't work. Some deps rely on the PHP version and would load a higher version.

greg0ire: Oh, I did not think of that. I think this happens because of the `minimum-stability` config entry, but cannot figure out why.

greg0ire: Can confirm, removing `minimum-stability` makes it work.

greg0ire:
Trying another trick.

greg0ire:
This time I think it is good!

composer.json
```diff
@@ -18,7 +18,10 @@
     ],
     "require": {
         "php": ">=5.4.0",
-        "symfony/symfony": "~2.2|~3.0",
+        "symfony/framework-bundle": "~2.2|~3.0",
+        "symfony/options-resolver": "~2.2|~3.0",
+        "symfony/twig-bridge": "~2.2|~3.0",
+        "sonata-project/admin-bundle": "~2.3|~2.4@dev",
```
core23: We should remove this asap when the ``SonataAdminBundle`` becomes final and/or we release the next version of this bundles.

greg0ire:
Ok so lesson of the day: don't use `minimum-dependencies` it opens too many possibilities for Composer.

composer.json
```diff
@@ -29,6 +32,9 @@
     "autoload": {
         "psr-4": { "Sonata\\TranslationBundle\\": "" }
     },
+    "config": {
+        "sort-packages": true
```
dbu: do we need this or was this just for debugging?

composer.json
```diff
@@ -18,8 +18,11 @@
     ],
     "require": {
         "php": ">=5.4.0",
-        "symfony/symfony": "~2.2|~3.0",
-        "sonata-project/core-bundle": "^2.2.1"
+        "sonata-project/admin-bundle": "~2.3|~2.4@dev",
```
dbu: i think we need to depend on admin-bundle actually. could we do ^2.3.0|~2.4@dev ? as sonata now respects semver, allowing a possible 2.5 version should not be a problem, right?

composer.json
```diff
@@ -18,8 +18,11 @@
...
+        "sonata-project/core-bundle": "^2.2.1",
+        "symfony/framework-bundle": "~2.2|~3.0",
+        "symfony/options-resolver": "~2.2|~3.0",
+        "symfony/twig-bridge": "~2.2|~3.0"
```
dbu: is this not already a dependency of sonata admin itself? same for the options resolver

composer.json
```diff
@@ -18,8 +18,11 @@
     ],
     "require": {
         "php": ">=5.4.0",
-        "symfony/symfony": "~2.2|~3.0",
-        "sonata-project/core-bundle": "^2.2.1"
+        "sonata-project/admin-bundle": "~2.3|~2.4@dev",
+        "sonata-project/core-bundle": "^2.2.1",
```
dbu: do we directly need the core bundle or only the admin, which in turn depends on core anyways?

composer.json
greg0ire: allowing 2.5 would be allowing non BC-breaking changes, so no, it would not be a problem IMO.

composer.json
greg0ire: @Soullivaneuh introduced this one, maybe the problem is not requiring the core, but making sure we get at least 2.1.1 ?

composer.json
greg0ire: It is neither needed nor for debugging. It is a best practice (makes sure dependencies are sorted alphabetically).

composer.json
greg0ire: Very very unrealistic, but what if a new version of the admin goes out, and it does not require this anymore. It would not be a BC-break. As soon as you use a package in the code, you must require it, no matter what your dependencies require IMO.

greg0ire: Looks like we do not reference the twig-bridge though (we do reference the option resolver).

greg0ire: But we are referencing twig itself, so there should be a dependency to twig.

composer.json
dbu: hm, sonata admin-bundle 2.3.0 requires `~2.2` so indeed this is a different requirement.

composer.json
dbu: @greg0ire can you please adjust this to `^2.3.0|~2.4@dev`?

composer.json
dbu: @greg0ire can you please adjust this to require twig itself then?

greg0ire: Ok I'll try to find the right requirement.

composer.json
greg0ire: I don't understand why I should do that. If we use semver, we want to be able to install 2.5 too, since it is supposed to be BC, don't you think ?

greg0ire: Oh sorry, I did not notice it was a caret, I'll adjust it as you wish.

dbu:
thanks @hason and @greg0ire !

PR #93, opened 2016-04-29
Add support for sonata-admin 3.x-dev

turko:
Allow 3.x-dev in composer

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -18,7 +18,7 @@
     ],
     "require": {
         "php": ">=5.4.0",
-        "sonata-project/admin-bundle": "^2.3.0|~2.4@dev|~3.x-dev@dev",
+        "sonata-project/admin-bundle": "^2.3.0|~2.4@dev|~3.x-dev",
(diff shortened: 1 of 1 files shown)


composer.json
```diff
@@ -18,7 +18,7 @@
     ],
     "require": {
         "php": ">=5.4.0",
-        "sonata-project/admin-bundle": "^2.3.0|~2.4@dev",
+        "sonata-project/admin-bundle": "^2.3.0|~2.4@dev|3.x-dev",
```
dbu: thanks for adressing this.

2.4 no longer exists. please use `^2.3.0|~3.0@dev`

core23: The current branch is ``3.x-dev@dev``

dbu: @core23 huh? isn't the point of aliasing the master branch that the requirement as i wrote it would work? then you would end up with the stable version once 3.0.0 is tagged.

Soullivaneuh:
Please see #94.

PR #94, opened 2016-05-01
Composer improvements

Soullivaneuh:
**THIS BUNDLE WILL HAVE A NEW MAJOR RELEASE**

* 2.x-dev
* Use new composer constraints notation

@dbu I would like to remove `1.0` branch and make a `1.x` instead do ou have new `1.0.x` tag to do before it?

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -17,17 +17,17 @@
         }
     ],
     "require": {
-        "php": ">=5.4.0",
-        "sonata-project/admin-bundle": "^2.3.0|~2.4@dev",
-        "sonata-project/core-bundle": "^2.2.1",
-        "symfony/framework-bundle": "~2.2|~3.0",
-        "symfony/options-resolver": "~2.2|~3.0",
-        "twig/twig": "~1.23"
+        "php": "^5.4 || ^7.0",
+        "sonata-project/admin-bundle": "^3.0",
+        "sonata-project/core-bundle": "^3.0",
+        "symfony/framework-bundle": "^2.3 || ^3.0",
+        "symfony/options-resolver": "^2.3 || ^3.0",
+        "twig/twig": "^1.23"
     },
     "require-dev": {
-        "knplabs/doctrine-behaviors": "~1.0",
-        "stof/doctrine-extensions-bundle": "~1.1",
-        "symfony/phpunit-bridge": "~2.7|~3.0",
+        "knplabs/doctrine-behaviors": "^1.0",
+        "stof/doctrine-extensions-bundle": "^1.1",
+        "symfony/phpunit-bridge": "^2.7 || ^3.0",
         "sllh/php-cs-fixer-styleci-bridge": "^2.0"
     },
     "autoload": {
@@ -37,14 +37,14 @@
         "sort-packages": true
     },
     "suggest": {
-        "knplabs/doctrine-behaviors": "~1.0",
-        "stof/doctrine-extensions-bundle": "~1.1",
-        "sonata-project/doctrine-orm-admin-bundle":"~2.2",
-        "sonata-project/doctrine-phpcr-admin-bundle": "~1.0"
+        "knplabs/doctrine-behaviors": "^1.0",
+        "stof/doctrine-extensions-bundle": "^1.1",
+        "sonata-project/doctrine-orm-admin-bundle":"^2.2",
+        "sonata-project/doctrine-phpcr-admin-bundle": "^1.0"
     },
     "extra": {
         "branch-alias": {
-            "dev-master": "1.1.x-dev"
+            "dev-master": "2.x-dev"
         }
     }
 }


composer.json
```diff
@@ -36,14 +36,14 @@
...
-        "sonata-project/doctrine-phpcr-admin-bundle": "~1.0"
+        "knplabs/doctrine-behaviors": "^1.0",
+        "stof/doctrine-extensions-bundle": "^1.1",
+        "sonata-project/doctrine-orm-admin-bundle":"^2.2",
+        "sonata-project/doctrine-phpcr-admin-bundle": "^1.0"
     },
     "extra": {
         "branch-alias": {
-            "dev-master": "1.1.x-dev"
+            "dev-master": "2.x-dev"
```
dbu: do we need to do BC breaking changes in the SonataTranslationBundle?

Soullivaneuh: As far as this project require `sonata-project/admin-bundle` and `sonata-project/core-bundle` yes, I'd prefer that way for a better start.

dbu:
why remove the 1.0 branch? it can be used if we need to do a maintenance release. but i just tagged 1.0.3 now so if you want to remove 1.0 and make master 1.x (aliased to 1.1 atm i would hope?) you can go ahead.

though i doubt if we need to bump the major version of this bundle - or do we need to do BC breaking changes to adjust to sonata?

Soullivaneuh:
> why remove the 1.0 branch? it can be used if we need to do a maintenance release.

To make all sonata repositories following same rules. :-)

See https://github.com/sonata-project/SonataAdminBundle/issues/3053

> or do we need to do BC breaking changes to adjust to sonata?

Yes this is a special case. We will try to strictly follow semver after that.

Soullivaneuh:
And of course, a `1.x` branch will be created instead. But it will be for history only just after the 2.0 stable release of this one.

dbu:
okay, go ahead then.

composer.json
```diff
@@ -17,17 +17,17 @@
...
     "require": {
-        "php": ">=5.4.0",
-        "sonata-project/admin-bundle": "^2.3.0|~2.4@dev",
-        "sonata-project/core-bundle": "^2.2.1",
-        "symfony/framework-bundle": "~2.2|~3.0",
-        "symfony/options-resolver": "~2.2|~3.0",
-        "twig/twig": "~1.23"
+        "php": "^5.4 || ^7.0",
+        "sonata-project/admin-bundle": "3.x-dev@dev",
+        "sonata-project/core-bundle": "3.x-dev@dev",
```
dbu: why do you write those dependencies like this? should it not be 3.*@dev? afaik that is the point of the 3.x-dev notation format. [the composer doc](https://getcomposer.org/doc/articles/aliases.md) has examples with 1.0.x-dev but i would expect that to work the same.

dbu:
note that the project can currently not be installed.

hason:
@Soullivaneuh, @dbu What is status? The bundle is not installable.

Soullivaneuh:
> @Soullivaneuh, @dbu What is status? The bundle is not installable.

What do you mean? Can't you use branch alias for now?

hason:
@Soullivaneuh No, because this package requires `"sonata-project/admin-bundle": "^2.3.0|~2.4@dev"` and `"sonata-project/core-bundle": "^2.2.1"`.

Soullivaneuh:
@hason Yes I know it. But did you try this? https://github.com/sonata-project/SonataAdminBundle/issues/3731#issuecomment-212636270

dbu:
@Soullivaneuh but that is really awkward. we should update composer.json here so that it works out of the box. i think in this PR the open question is why we do not require `^3.0@dev` but instead this `3.x-dev@dev` which is double strange (@ dev is not needed in a -dev branch requirement, and the point of the branch alias is that you can write `3.0@dev`)

WouterJ:
Can we please merge this? (or at least, move this forward)

Soullivaneuh:
@WouterJ I working on it. I hope to release a new version this week.

dbu:
created the 1.x branch and tagged [1.1.0](https://github.com/sonata-project/SonataTranslationBundle/releases/tag/1.1.0)

deleting the 1.0 branch now

Soullivaneuh:
@dbu Why merging this? this was not finished nor reviewed.

Please wait a bit before tagging.

dbu:
i won't tag a 2.0, no worries. the diff looked ok to me, thats why i merged.

not sure what sonata policy is, i got very used to have a [WIP] in the PR title when something is not for merge yet.

dbu:
the 1.1.0 is tagged on 1.x branch which does not contain this commit, as this starts the 2.0 version...

composer.json
```diff
@@ -37,14 +37,14 @@
         "sort-packages": true
     },
     "suggest": {
-        "knplabs/doctrine-behaviors": "~1.0",
-        "stof/doctrine-extensions-bundle": "~1.1",
-        "sonata-project/doctrine-orm-admin-bundle":"~2.2",
-        "sonata-project/doctrine-phpcr-admin-bundle": "~1.0"
+        "knplabs/doctrine-behaviors": "^1.0",
+        "stof/doctrine-extensions-bundle": "^1.1",
+        "sonata-project/doctrine-orm-admin-bundle":"^2.2",
```
Soullivaneuh: Note: Must be `^3.0`.

composer.json
```diff
@@ -37,14 +37,14 @@
...
+        "sonata-project/doctrine-phpcr-admin-bundle": "^1.0"
```
Soullivaneuh: Note: Check packge status before tag translation-bundle.

composer.json
dbu: will you follow up on this one?

Soullivaneuh: Of course! :+1:

composer.json
Soullivaneuh: @dbu We have an issue here.

I have to push a 2.0 version of `sonata-project/doctrine-phpcr-admin-bundle` before this one but we hare stuck because of this: https://github.com/sonata-project/SonataDoctrinePhpcrAdminBundle/issues/373

Any suggestion? We can not use `^1.0` because this one is not compatible with admin-bundle `3.0`.

WouterJ: @Soullivaneuh this bundle only suggests the PHPCR bundle, it does not require it. Using versions is suggestions doesn't do anything.

For now, I think the best option is to release this bundle 2.0 and then require people using the PHPCR bundle to use 2.0@dev (which, due to the strange branch aliases used in Sonata, has to be `2.*@dev`). These people are probably already using Symfony CMF, so they need lots of dev versions in order to use Sonata 3.0.

Soullivaneuh: > @Soullivaneuh this bundle only suggests the PHPCR bundle, it does not require it. Using versions is suggestions doesn't do anything.

If it suggested, this mean piece of code managing this bundle is present here, right? Is that not tested?

WouterJ: If I'm correct (I know almost nothing about this bundle), the bundle just comes with a PHPCR implementation. As you need Sonata PHPCR adapter when using  PHPCR, it suggests you to check that bundle.

It doesn't depend on the PHPCR adapter directly anywhere afaik.

dbu: SonataTranslationBundle integrates with both orm admin and phpcr admin. indeed i think our best solution is to suggest 2.0@dev (or whatever we need to do if the master branch alias is strange) and move on. the phpcr integration is not particularly deep, so i don't expect that we need to change things once phpcr admin 2.0 is released. and indeed, cmf users should not migrate to sonata 3 yet, too many things not ready.

Soullivaneuh: @WouterJ @dbu So:

* Add a conflict key for `sonata-project/doctrine-phpcr-admin-bundle` `<2.0`
* Integrate this project on dev-kit
* Prepare it for 2.0.0 (changelog, upgrade etc...)
* Push the release

Agree with that?

Soullivaneuh: >  and indeed, cmf users should not migrate to sonata 3 yet, too many things not ready.

This is also the problem. If I release translation-bundle 2.0 compatible only with sonata-admin 3, the 1.x branch will be considered as legacy and should not be maintained anymore.

The code should be stable, but it would be great to have a full compatibility soon.

WouterJ: :-1: on the conflict key:

 * This isn't done for other suggested packages
 * <2.0 will never be installed with 2.0 of this bundle, as they both require another major of SonataAdminbundle
 * Conflict keys aren't nice in Composer AFAIK

Integrating on dev-kit, your choice, you're the dev-kit master here :)

And I can't say any significant about releasing a major of this bundle. I'm not [very familair with this bundle](https://github.com/sonata-project/SonataTranslationBundle/commits?author=WouterJ).

Soullivaneuh: > This isn't done for other suggested packages

Yeah and it's a mistake. The `suggest` should contains a description of why this package is suggested, not a version constraint.

The `conflict` key help to ensure compatibility when requiring the suggested package.

So this have to be changed on all other repository, let's start now. :+1: 

> <2.0 will never be installed with 2.0 of this bundle, as they both require another major of SonataAdminbundle

Didn't understand. Could you please elaborate?

> Conflict keys aren't nice in Composer AFAIK

From where do you know that?

greg0ire: > Conflict keys aren't nice in Composer AFAIK

You mean until proven nice or you have something to back that statement? The only nasty thing that I see is that if you use the suggested package for something else than integration with the bundle, you cannot use a conflicting version, but that is also true of any required package.

WouterJ: > Didn't understand. Could you please elaborate?

 * SonataDoctrinePhpcrAdminBundle 1.x requires SonataAdminBundle 2.x (it doesn't support 3.x).
 * SonataTranslationBundle 2.x requires SonataAdminBundle 3.x (it doesn't support 2.x)

This means that SonataDoctrinePhpcrAdminBundle 1.x and SonataTranslationBundle 2.x will never be installed at the same time. So adding a conflict rule for SonataDoctrinePhpcrAdminBundle <2.0 in SonataTranslationBundle 2.x is redundant and not needed (if anything, it would probably make version resolving complexer because there are more constraints).

 > From where do you know that?

I remember reading that conflict rules are check at a different position then the require version constraints (I believe after). That makes conflict rules less efficient. I'm now looking for more concrete information to back this statement though.

greg0ire: PR for fixing suggest : #100

Soullivaneuh: > SonataDoctrinePhpcrAdminBundle 1.x requires SonataAdminBundle 2.x (it doesn't support 3.x).

This have to be refactored too. :+1:

greg0ire: @WouterJ could you also explain what's wrong exactly with our branch aliases ?

greg0ire: > This means that SonataDoctrinePhpcrAdminBundle 1.x and SonataTranslationBundle 2.x will never be installed at the same time.

That's right, I changed #100 to reflect that.

WouterJ: > This have to be refactored too. :+1:

SonataDoctrinPhpcrAdminBundle 2.x will support 3.x, so I don't think we have to refactor anything.

 > could you also explain what's wrong exactly with our branch aliases ?

You're using `2.x-dev`, while it's more common to use `2.0.0-dev` (or `2.0.x-dev`). For some reason, this little difference means that it won't install the `2.x-dev` branch when using `2.0.*@dev` as requirement. You have to explicitly say `^2.0@dev` (or `~2.0@dev`).

Because the merged PRs determine whether the next version will be a patch or minor release, it's impossible to for instance create a good requirement for the `3.x` branch in SonataAdminBundle. (`3.*@dev` won't work, `^3.2@dev` might work, unless the next release is going to be 3.1.1, in which case `^3.1.1@dev` had to be used).

Soullivaneuh: @WouterJ Are you informed about https://github.com/sonata-project/SonataAdminBundle/issues/3053 https://github.com/sonata-project/SonataAdminBundle/issues/3731

In a nutshell, we did 2.x, 3.x etc branches because we don't have enough resources to maintain minor branch.

We did it before, see the result on admin-bundle...

> SonataDoctrinPhpcrAdminBundle 2.x will support 3.x, so I don't think we have to refactor anything.

Yes, I started it: https://github.com/sonata-project/SonataDoctrinePhpcrAdminBundle/commit/dfcb64ce249f68fe27409b799e2e8ebb88b58c38

I'm talking about also dev-kit integration and stuff to start the new release management on this sonata project.

WouterJ: > Are you informed about

Yes, I'm aware of that and I fully agree with this discussion. (less resources is a big problem in the Symfony world at the moment and this plan at least gave Sonata project development a big boost the last months, big :+1: to you guys!)
@greg0ire  just asked me to explain what's weird about the branch alias and what problems it caused, so I explained them.

It would be nice to know if the master branch is going to be a patch or minor release though. That would solve a big part of the issue.

Soullivaneuh: The master branch will be a new major release.

PR #95, opened 2016-05-02
StyleCI integration

Soullivaneuh:


Makefile
```diff
@@ -1,9 +1,3 @@
-cs:
-	php-cs-fixer fix --verbose
```
OskarStark: so there is no way to fix it locally? i nee to push and use Style CI?

dbu:
:+1: i let you merge @Soullivaneuh once you discussed the question with @OskarStark

Makefile
Soullivaneuh: You can always use php-cs-fixer. But having a command not used anymore by travis does not make sens.

BTW, not all of the people has php-cs-fixer globally installed. Some can have a `.phar` and the `make cs` command will be useless.

OskarStark: ok fine for me!

push the button

PR #109, opened 2016-06-09
Remove doctrine phpcr admin conflict

WouterJ:
As I already tried to explain, this bundle **does not depend on SonataDoctrinePhpcrOdmAdminBundle**. It makes absolutely no sense to set up conflict rules for this package.

A proof that this bundle doesn't use anything from that bundle: https://github.com/sonata-project/SonataTranslationBundle/search?utf8=%E2%9C%93&q=DoctrinePhpcr

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -39,7 +39,6 @@
     "conflict": {
         "knplabs/doctrine-behaviors": "<1.0 || >=2.0",
         "sonata-project/doctrine-orm-admin-bundle": "<3.0 || >=4.0",
-        "sonata-project/doctrine-phpcr-admin-bundle": "<2.0 || >=3.0",
         "stof/doctrine-extensions-bundle": "<1.1 || >=2.0"
     },
     "autoload": {


greg0ire:
I moved this to the conflict section the other day, but you're right, I can't see why there would be a need for that in the first place.

OskarStark:
But if you search for this: https://github.com/sonata-project/SonataTranslationBundle/search?utf8=%E2%9C%93&q=Phpcr&type=Code

you find something, can we please double check that?

greg0ire:
Oh wait, what about this search : https://github.com/sonata-project/SonataTranslationBundle/search?utf8=%E2%9C%93&q=phpcr ?

OskarStark:
haha

greg0ire:
So if there should be a conflict, it should be with the Doctrine PHPCr package itself, no ?

OskarStark:
> As you can see, no service or class of the SonataDoctrinePhpcrAdminBundle is used (precisely why it isn't a requirement).

thank you for clarify this @WouterJ 👍

Soullivaneuh:
> As you can see, this bundle has PHPCR support.

And we must define **which version** we support.

Maybe no code is here because the support **is not tested** yet, isn't it?

This should be reverted.

greg0ire:
> And we must define which version we support.

Of PHPCR, not of the bundle, right ?

WouterJ:
> Maybe no code is here because the support is not tested yet, isn't it?

Given that we use it in the CMF project (1.3 and 2.0-dev), I think it's safe to say that everything is tested. But indeed, proper functional tests are missing for PHPCR in this bundle.

 > This should be reverted.

I think I'm not very good at explaining or the like. This bundle and SonataDoctrinePhpcrAdminBundle have **no relation** with eachother, except that they both are maintained by the Sonata project.

As you can see, ORM is proper tested, but this bundle isn't used at all either by this package. Just because they don't have any relation. This means that this conflict should be removed as well.

At the moment, you support `doctrine/phpcr-odm` `^1.0` (which is just all PHPCR ODM versions). You might want to specify that as conflict, but I don't see much reason to do so.

PR #110, opened 2016-06-09
Remove ORM admin bundle conflict

WouterJ:
Same as #109

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -38,7 +38,6 @@
     },
     "conflict": {
         "knplabs/doctrine-behaviors": "<1.0 || >=2.0",
-        "sonata-project/doctrine-orm-admin-bundle": "<3.0 || >=4.0",
         "stof/doctrine-extensions-bundle": "<1.1 || >=2.0"
     },
     "autoload": {


Soullivaneuh:
:-1: For this one.

If this suggested, we **have** to specify for which version.

WouterJ:
This bundle **doesn't use** this package. It makes no sense to conflict with something you don't use.

You might as well just add `"laravel/laravel": "^5.0"` to the conflict rule.

Soullivaneuh:
So why this package is on the suggest section from long time ago? Oo

WouterJ:
Because, if you're going to need ORM support for your i18n flags, you might as well need ORM support for your admin panel. This is why it also suggests to look over that.

However, it doesn't depend on it as it works on a complete other level. I think it might make sense to read this bundle's code, its documentation and try it out in a project. To be honest, I have the feeling that you're making decisions without knowing exactly what this bundle is doing.

Soullivaneuh:
> To be honest, I have the feeling that you're making decisions without knowing exactly what this bundle is doing.

This is why I was asking:

> So why this package is on the suggest section from long time ago?

On most of the Sonata Projects, some package are suggested **because** it would be use by the project if they are require.

Also, packages on suggest was specifying a version too. That is first, not right AFAIK, second seems to indicate a minimum version to resolve.

So this make sense to me to move them on the `conflict` key, because of version specification, that's all.

If this should not be on conflicted, I don't see why a version was specify on `suggest` section. Are we agree at least for this point?

WouterJ:
> If this should not be on conflicted, I don't see why a version was specify on suggest section. Do we agree at least for this point?

Well, suggested packages are used for 2 things:

 1. Packages with optional integrations in the package
 2. Packages that are good to check out as well

Both Sonata adapter packages are suggested because of reason (2): It makes no sense to use this bundle without either the ORM or PHPCR adapter bundle (and this bundle doesn't support MongoDB documents, so it doesn't suggest that adapter bundle).

If you've decided that you only want packages in the suggests section because of (1), you should indeed remove this both suggestions from the composer.json.

Soullivaneuh:
> If you've decided that you only want packages in the suggests section because of (1), you should indeed remove this both suggestions from the composer.json.

In this case, I fully :+1: your PR. :wink: 

I think this can indeed be removed from suggestions too.

This bundle requires `sonata-project/admin-bundle` that need a Persistency layer. So the suggestion is implicit.

Could please update this? Thanks.

Soullivaneuh:
Same for `sonata-project/doctrine-phpcr-admin-bundle` in this case.

WouterJ:
Created https://github.com/sonata-project/SonataTranslationBundle/pull/111 as it was easier to do with the GitHub GUI :grin:

PR #117, opened 2016-07-05
Bump the PHPCR ODM conflict

WouterJ:
This bundle currently supports Doctrine PHPCR ODM 2. This is needed to get the CMF sandbox installable again /cc @dbu

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -36,7 +36,7 @@
         "stof/doctrine-extensions-bundle": "if you translate orm entities with the gedmo extensions"
     },
     "conflict": {
-        "doctrine/phpcr-odm": ">=2.0",
+        "doctrine/phpcr-odm": ">=3.0",
         "knplabs/doctrine-behaviors": "<1.0 || >=2.0",
         "stof/doctrine-extensions-bundle": "<1.1 || >=2.0"
     },


Soullivaneuh:
> This bundle currently supports Doctrine PHPCR ODM 2

Is this stable? Because if not this version will never be installed on Travis, so no tested.

So no real guarantee this bundle is compatible with this version.

WouterJ:
No, it's not yet stable. But please keep this in mind:

 * There are no PHPCR tests in this bundle at all, so whether it can or can't be installed on Travis is not related
 * This bundle only uses `findTranslation`/`bindTranslation`, these are such important part of the main API of Doctrine that there won't be any BC break in these methods.

WouterJ:
> The real question is: Why this is not tested? 😉

I don't know, I've only contributed composer.json changes to this project.

 > You mean doctrine/phpcr-odm use translation-bundle methods? In this case this should not be on this compose.json at all, should it be?

No, PHPCR ODM has built in i18n support. This bundle only integrates it into sonata by calling find/bind translations on PHPCR ODM documents.

Soullivaneuh:
Oh okay. So IMHO it should be tested. Implemented test or not yet, doing this is risky.

WouterJ:
> Oh okay. So IMHO it should be tested. Implemented test or not yet, doing this is risky.

I agree. Same applies to Knp integration btw. Afaics, only gedmo is tested

dbu:
> IMHO it should be tested. Implemented test or not yet, doing this is risky.

agreed. but no idea if we find a volunteer to do it...

almacbe:
@dbu @WouterJ @Soullivaneuh I could try. At least for Knp integration. If I have some doubt I will ask you. If you give the first ideas, steps or how to start I do it.

dbu:
@almacbe sounds good!

currently, the tests for gedmo only test the models provided in this bundle. imho, to tell whether we really are compatible with a version of doctrine, we would need functional tests. that would mean to set up doctrine orm with the knp (resp. gedmo) extension in a test case. simplest is to use sqlite for that to avoid having to set up a more complicated datatabse.

with functional tests, we could also test if the admin extension really works with sonata as expected.
@Soullivaneuh do we have some base setup for functional tests in sonata bundles? if not, this could be used as inspiration: https://github.com/FriendsOfSymfony/FOSHttpCacheBundle/tree/master/Tests/Functional/Fixtures

PR #120, opened 2016-08-13
Update translation-bundle version installation.rst

NDFA:
Documentation files are read from this branch.

## Changelog

```markdown
### Added
- Added some `Class::newMethod` to do great stuff

### Changed

### Deprecated

### Removed

### Fixed

### Security
```

## To do

- [ ] Update the tests
- [ ] Update the documentation
- [ ] Add an upgrade note

## Subject

Update translation-bundle version for composer in installation.rst

### DIFF
--- a/Resources/doc/reference/installation.rst
+++ b/Resources/doc/reference/installation.rst
@@ -14,6 +14,20 @@ following command to download the latest stable version of this bundle:
 This command requires you to have Composer installed globally, as explained
 in the `installation chapter`_ of the Composer documentation.
 
+If you are facing composer errors because of dependency miss-match like:
+
+.. code-block:: bash
+    
+    Your requirements could not be resolved to an installable set of packages.
+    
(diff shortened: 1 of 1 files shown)


core23:
:-1: this is not related to this bundle, but general composer behavior

greg0ire:
Sorry, but this is indeed off topic. Thanks for trying to help though.

NDFA:
OK, But there is another issue like this (#61) which arises coz of dependent bundle version incompatibility. I think its better to reduce interruptions occurring during following documentation to Google errors.

greg0ire:
This would be better in stack overflow, because it is a q and a site

dbu:
i think what should be done is simply to remove the version with `composer require` in the doc. composer then picks the latest one, which is more convenient.

greg0ire:
:+1: can you contribute that @NDFA?

NDFA:
Sure Dude :-)

PR #129, opened 2016-08-29
Display flags for locales with country code

dmarkowicz:
I am targetting this branch, because it is fix for 2.x.

## Changelog

```markdown
### Fixed
- Display flags for locales with country code
```

## Subject
Locale Switcher Block does not display flag icons for locales with country codes, i.e.: pt_BR, en_US. There are no such images, so user see 404 instead. This solution just take last 2 characters from locale and display proper icon. It will be:
`pt.png` for `pt`
`br.png` for `pt_BR`
`us.png` for `en_US`
etc.


OskarStark:
Can you please add a screenshot before/after ?

dmarkowicz:
For testing purposes I added locale label next to flag.

Before patch:
locale-switcher

After patch:
locale-switcher-patched

OskarStark:
Thank you @dmarkowicz 👍

PR #132, opened 2016-09-16
Duplicate jp flag to ja to fit iso locale code

nvbooster:
## Changelog
```markdown
### Added
- Addded missing flag image `ja.png`
```

## Subject
Wrong locale icon in locale switcher for japanese locale

File name: jp.png
Locale name: ja

## Steps to reproduce
sonata_translation:
    locales: [en, ja]
    default_locale: en
## Expected results
In top of admin GB and Japanese flags
## Actual results
Japanese flag icon has tag

``

which cannot be load  (as only jp.png presents in bundle resources)


nvbooster:
sonata-project/SonataTranslationBundle#132

OskarStark:
Do we still need the `jp.png` then @nvbooster or could it be removed?

nvbooster:
I think its still required, because of 

``

in template. It is valuable in case full locale name "ja_JP" is used

OskarStark:
sounds reasonable to me 👍

OskarStark:
ping @sonata-project/contributors

greg0ire:
This fix does not seem right. If we interchangeably use a country code and a language code, something is really not right. So what is this flag about? A country, or a language?

dbu:
locales are both countries and languages. according to ISO, ja stands for the language, jp the country. i would keep both (if we remove jp, it can only be in a major version - people might have worked around the mistake and rely on having the jp one). 

`locale[-2:2]|lower ~ '.png'` sounds seriously weird though. the second part of the xx_XX code is always the country. it would mean we would also need a ch.png for switzerland. but in switzerland people speak french and german and italian, so de_CH, fr_CH and it_CH are all valid things. so "ch" is definitely not a language but a country. it feels like the flag should fall back to the language, not the country when a long code is used. if anything, we could offer images for de_CH (or more likely, things like pt_BR for brasilian portuguese).

i suggest we just add ja.png here, and if somebody can solve the other issue, do that in a separate pull request.

OskarStark:
Thanks for the detailed feedback @dbu 👍  and thank you for the contribution @nvbooster 👍

greg0ire:
Fine, let's do that. But this raises a lot of questions, I wonder what is right way to do this… for instance, for `fr_CA`, I would put the flag of Québec.

dbu:
created #133 to further discuss the general issue. i hope there is way to overwrite icons? then people can choose whether they want a split french+canadian flag icon, or a quebec icon.

PR #157, opened 2017-03-11
Fix section headers in orm doc

dmarkowicz:
I am targeting this branch, because it is documentation fix.

## Changelog

```markdown
### Changed
updated documentation
```
## Subject

As @greg0ire noticed in https://github.com/sonata-project/SonataTranslationBundle/issues/154#issuecomment-285636156 the numbering in orm doc could be improved. I removed manual numbers and fixed hierarchy.

PR #160, opened 2017-03-15
fixed headlines in docs

### LINKED ISSUE #159: documentation title numerotation issue
There is an issue on your markdown reader.
The title's hierarchy isn't respected https://sonata-project.org/bundles/translation/master/doc/reference/orm.html but it seems ok on the github preview https://github.com/sonata-project/SonataTranslationBundle/blob/2.x/Resources/doc/reference/orm.rst

OskarStark:
I am targeting this branch, because this is a doc fix.

Closes #159

### DIFF
--- a/Resources/doc/reference/orm.rst
+++ b/Resources/doc/reference/orm.rst
@@ -2,8 +2,12 @@
 Translate Doctrine ORM models
 =============================
 
-Scenario A - Using Gedmo Doctrine Extensions
-============================================
+You can either use :ref:`Gedmo Doctrine Extensions <gedmo_doctrine_extensions>` or :ref:`KnpLabs Doctrine Behaviours <knp_labs_doctrine_bahaviours>`.
+
+.. _gedmo_doctrine_extensions:
+
+Using Gedmo Doctrine Extensions
+===============================
 
 Doctrine ORM models translations are handled by `Gedmo translatable extension`_.
 
@@ -208,8 +212,10 @@ Example for configure search filter
                 ->add('title', TranslationFieldFilter::class); // or 'doctrine_orm_translation_field'
         }
 
-Scenario B - Using KnpLabs Doctrine Behaviours
-==============================================
+.. _knp_labs_doctrine_bahaviours:
+
+Using KnpLabs Doctrine Behaviours
+=================================
 
 Implement TranslatableInterface
 -------------------------------


greg0ire:
- "numeration" is not a word
- keep it under 50 chars please

OskarStark:
> "numeration" is not a word

fixed

> keep it under 50 chars please

fixed

Resources/doc/reference/orm.rst
```diff
@@ -2,8 +2,8 @@
 Translate Doctrine ORM models
 =============================
 
-Scenario A - Using Gedmo Doctrine Extensions
-============================================
+Option A - Using Gedmo Doctrine Extensions
+==========================================
```
greg0ire: @sstok suggest that this might be b/c we use `=` here, just like on line 3. Can you try reverting to Scenario, and use something else than `=` for subtitles?

OskarStark: To be honest, i don't really understand what you mean?

greg0ire: I mean changing all the `=` line 6 to something like `~` or `-`

OskarStark: ok, I will try, but anyway, it makes no sense, that `Scenario` disappears

OskarStark: Based on this: http://thomas-cokelaer.info/tutorials/sphinx/rest_syntax.html#headings

```
Normally, there are no heading levels assigned to certain characters as the structure is determined from the succession of headings. However, it is better to stick to the same convention throughout a project. For instance:

# with overline, for parts
* with overline, for chapters
=, for sections
-, for subsections
^, for subsubsections
“, for paragraphs
```

`=` is a section, and therefore correct

greg0ire: if `=` is a section then everything under line 3 is and it means we are dealing with a subsection => you should use `-`

greg0ire: Or alternatively, change line 3 with `*`

OskarStark:
I changed the first headline and used `#` for underlining

Looks good to me:
screenshot 2017-03-16 09 03 05

OskarStark:
Hmm, but all our other doc files starting with a headline like this:

```
First Headline
==========
```

let me propose something else

OskarStark:
I removed the `Scenario`-part from the headline and keep the correct headline hierachy.
I also added an introduction sentence with acnhor links down to the two available options

Resources/doc/reference/orm.rst
```diff
@@ -2,8 +2,12 @@
...
 =============================
 
-Scenario A - Using Gedmo Doctrine Extensions
-============================================
+You can either use :ref:`Gedmo Doctrine Extensions <gedmo_doctrine_extensions>` or :ref:`KnpLabs Doctrine Behaviours <knp_labs_doctrine_bahaviours>`.
+
+.. _gedmo_doctrine_extensions:
+
+Using Gedmo Doctrine Extensions
+===============================
```
greg0ire: Github properly uses an h2 here, although you are still using the same sign. I don't get how rst works, but maybe this is bug-based :confused:

OskarStark: The symfony docs for example, all the time start like this for the first headline, normally we have to change all headlines. Anyway I would like to rework all the docs and bring them in common. But for now the numbering is fixed

greg0ire:
Ok let's go with that then! Thanks!

Resources/doc/reference/orm.rst
wouterj: reStructured text allows the following characters to be used as title underline: ``! " # $ % & ' ( ) * + , - . / : ; < = > ? @ [ \ ] ^ _ ` { | } ~``. From this big list, the recommended ones by reSt are: ``= - ` : . ' " ~ ^ _ * + #``

The order in which you use them isn't important. When a new underline char is found, the level will simply be increased by one (h2 -> h3). This hierarchy will then be saved and used throughout that *document*.

For consistency, the [Symfony doc standards](http://symfony.com/doc/current/contributing/documentation/standards) defines this rule: *"The following characters are chosen for different heading levels: level 1 is = (equal sign), level 2 - (dash), level 3 ~ (tilde), level 4 . (dot) and level 5 " (double quote);"*

greg0ire: > The order in which you use them isn't important. When a new underline char is found, the level will simply be increased by one (h2 -> h3). This hierarchy will then be saved and used throughout that document.

@wouterj I knew  that part, what I don't understand is why an `h2` is used although the same char is used. 

> normally we have to change all headlines. Anyway I would like to rework all the docs and bring them in common

I think we should rework the docs too, in one big PR. Thank you for proposing @OskarStark 

> For consistency, the Symfony doc standards defines this rule: "The following characters are chosen for different heading levels: level 1 is = (equal sign), level 2 - (dash), level 3 ~ (tilde), level 4 . (dot) and level 5 " (double quote);"

We should do the same so that it looks familiar to sf docs contributors IMO.

wouterj: > @wouterj I knew that part, what I don't understand is why an h2 is used although the same char is used.

I think that's because *"As a general rule, a document can only have one top level headline."*, however I couldn't find it this quickly in the reSt specs.

greg0ire: You're probably right. If it isn't in the specs, we should probably stay away from it though.

OskarStark: Thanks for the detailed feedback @wouterj 👍 

> I think we should rework the docs too, in one big PR. Thank you for proposing @OskarStark

yes, will do this soon!

> For consistency, the Symfony doc standards defines this rule: "The following characters are chosen for different heading levels: level 1 is = (equal sign), level 2 - (dash), level 3 ~ (tilde), level 4 . (dot) and level 5 " (double quote);"
> 
> As a general rule, a document can only have one top level headline.

@greg0ire an you put this general rule (which should be applied to all new docs from now on) to the `CONTRIBUTING.md`?
And *yes* we should follow these symfony-docs rules too

greg0ire: > @greg0ire an you put this general rule (which should be applied to all new docs from now on) to the CONTRIBUTING.md?

Adding that to my todo list

dmarkowicz: I added overline = to first headline in https://github.com/sonata-project/SonataTranslationBundle/pull/157/files#diff-046a0b7c44e8854e46ceffe912928724R1 to make it H1 without rewriting all other headlines below. According to http://docutils.sourceforge.net/docs/ref/rst/restructuredtext.html#sections. This probably caused some confusion 😳
```
== <----- 
H1
==

H2
==
```
#159 refers to an outdated version of the documentation from master branch which is not merged with 3.x.

greg0ire: Oh right I didn't even notice it

greg0ire: I added that contribution to [a PR I already have](https://github.com/sonata-project/dev-kit/pull/203/files#diff-55aa24e78df8010aea154e9506cab16dR97)

PR #185, opened 2017-09-28
Fix RuntimeException in sonata:admin:explain command

### LINKED ISSUE #164: 'The Request object has not been set' with 'console sonata:admin:explain ...'
## Steps to reproduce

Excecute :

` console sonata:admin:explain ` _service_

## Actual results

>  [RuntimeException]
>  The Request object has not been set

## Solution 

In file "/Admin/Extension/AbstractTranslatableAdminExtension.php" line 71 replace 
```diff
- if ($admin->getRequest()) {
+ if ($admin->hasRequest()) {
    $this->translatableLocale = $admin->getRequest()->get(self::TRANSLATABLE_LOCALE_PARAMETER);
 }
```

dmarkowicz:
I am targeting this branch, because it is bugfix.

Closes #164

## Changelog

```markdown
### Fixed
- Fix RuntimeException in sonata:admin:explain command
```

## Subject
RuntimeException is thrown in `sonata:admin:explain` command and output is truncated.

### DIFF
--- a/Admin/Extension/AbstractTranslatableAdminExtension.php
+++ b/Admin/Extension/AbstractTranslatableAdminExtension.php
@@ -68,7 +68,7 @@ abstract class AbstractTranslatableAdminExtension extends AbstractAdminExtension
     public function getTranslatableLocale(AdminInterface $admin)
     {
         if ($this->translatableLocale == null) {
-            if ($admin->getRequest()) {
+            if ($admin->hasRequest()) {
                 $this->translatableLocale = $admin->getRequest()->get(self::TRANSLATABLE_LOCALE_PARAMETER);
             }
             if ($this->translatableLocale == null) {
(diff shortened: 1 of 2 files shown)


Admin/Extension/AbstractTranslatableAdminExtension.php
```diff
@@ -68,7 +68,7 @@ public function getTranslatableChecker()
     public function getTranslatableLocale(AdminInterface $admin)
     {
         if ($this->translatableLocale == null) {
-            if ($admin->getRequest()) {
+            if ($admin->hasRequest()) {
```
core23: This looks right, but I'm wondering how this exception could occur. `hasRequest` is just a simple wrapper for a null check.

dmarkowicz: There is no HTTP request in a CLI command.

core23: I understand the problem, but I don't understand why this could happen...

`if ($admin->hasRequest()) {` 

... would be evaluated to ...

`if ($admin->getRequest() !== null) {`

... which should be the same like ...

`if ($admin->getRequest()) {`

core23:
Thanks @dmarkowicz

PR #210, opened 2017-12-08
Rework Symfony dependencies + SF4 compat

jordisala1991:
I am targeting this branch, because this is BC.

## Changelog

```markdown
### Added
- Compatibility with Symfony4
```

## Subject

Remove direct dependency with framework bundle (only needed for test), and make Symfony dependencies explicit.

Also allow symfony4 to be installed

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -21,12 +21,12 @@
         "sonata-project/admin-bundle": "^3.28",
         "sonata-project/block-bundle": "^3.8",
         "sonata-project/core-bundle": "^3.7.1",
-        "symfony/config": "^2.8 || ^3.2",
-        "symfony/dependency-injection": "^2.8 || ^3.2",
-        "symfony/form": "^2.8 || ^3.2",
-        "symfony/http-foundation": "^2.8 || ^3.2",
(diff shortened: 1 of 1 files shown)


composer.json
```diff
@@ -31,7 +35,8 @@
         "matthiasnoback/symfony-dependency-injection-test": "^1.0",
         "sonata-project/doctrine-orm-admin-bundle": "^3.2",
         "stof/doctrine-extensions-bundle": "^1.1",
-        "symfony/phpunit-bridge": "^3.3"
+        "symfony/framework-bundle": "^2.8 || ^3.2 || ^4.0",
+        "symfony/phpunit-bridge": "^3.3 || ^4.0"
```
core23: Can we drop 3.3 here?

jordisala1991: Would be better to do it when we have a reason. when we add the listener on the phpunit.xml.dist for example

OskarStark: this will be done by dev-kit in the near future

jordisala1991: Then when dev-kit creates the PR we modify it changing this version

PR #225, opened 2018-02-21
Improve docs for importing css

kunicmarko20:
I am targeting this branch, because I am improving docs.

## Subject

Adding asset to `sonata_admin` config looks a lot easier to me than extending the template and replacing it.

### DIFF
--- a/docs/reference/installation.rst
+++ b/docs/reference/installation.rst
@@ -99,30 +99,23 @@ Key                 Description
 Step 4: Import the Styles
 -------------------------
 
-Extend the `SonataAdminBundle layout`_ and add the SonataTranslationBundle stylesheet like this:
+Install SonataTranslationBundle web assets under your public web directory:
 
-.. code-block:: html+jinja
+.. code-block:: bash
 
-    {# templates/admin/layout.html.twig #}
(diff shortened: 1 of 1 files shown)


docs/reference/installation.rst
```diff
@@ -99,30 +99,21 @@ Key                 Description
 Step 4: Import the Styles
 -------------------------
 
-Extend the `SonataAdminBundle layout`_ and add the SonataTranslationBundle stylesheet like this:
+Installing assets as hard copies:
```
kunicmarko20: `Install assets as hard copies`

docs/reference/installation.rst
```diff
@@ -99,30 +99,23 @@ Key                 Description
 Step 4: Import the Styles
 -------------------------
 
-Extend the `SonataAdminBundle layout`_ and add the SonataTranslationBundle stylesheet like this:
+Install assets as hard copies:
```
greg0ire: I think there's a config entry right in Composer.json to choose between copies, relative symlinks or absolute symlinks, so this might not be true

kunicmarko20: @greg0ire is this better? (I copied commands description directly)

greg0ire:
Thanks @kunicmarko20 !

PR #229, opened 2018-04-20
minor fix orm.rst

fbertolotti:


kunicmarko20:
Not sure if this should be fixed, I think we use `\` everywhere.

dbu:
afaik in .rst source code is double backtick (the ` character). if you use those, the single backslash should be fine.

kunicmarko20:
Just checked rest of docs and you are correct, @fbertolotti could you change it to double backtick and with jus tone backslash.

fbertolotti:
I've updated it :)

greg0ire:
To be squash merged

greg0ire:
Thanks @fbertolotti !

PR #241, opened 2018-09-26
Make TranslationFieldFilter respect condition

### LINKED ISSUE #240: TranslationFieldFilter does not respect setCondition
## Subject

When using GlobalSearch, filters are built with OR condition:

```php
if ($filter->getOption('global_search', false)) {
    $filter->setCondition(FilterInterface::CONDITION_OR);
    $datagrid->setValue($filter->getFormName(), null, $term);
    $found = true;
 }
```

however, TranslationFieldFilter applies andWhere explicitly: https://github.com/sonata-project/SonataTranslationBundle/blob/2.x/src/Filter/TranslationFieldFilter.php#L56

It should respect $condition and act accordingly. Should I PR that?

andrewtch:
Make TranslationFieldFilter respect condition, fixes global search on translatable fields. Fixes #240

### DIFF
--- a/src/Filter/TranslationFieldFilter.php
+++ b/src/Filter/TranslationFieldFilter.php
@@ -53,7 +53,7 @@ final class TranslationFieldFilter extends Filter
         }
 
         // search on translation OR on normal field
-        $queryBuilder->andWhere($queryBuilder->expr()->orX(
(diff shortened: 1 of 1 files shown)


greg0ire:
Thanks @andrewtch

PR #245, opened 2018-11-14
Applied fixes from FlintCI

Soullivaneuh:


core23:
We should require the latest phpunit bridge to fix this

OskarStark:
Can you provide a PR ?

PR #274, opened 2019-01-13
[Docs] big cleanup

OskarStark:


docs/reference/orm.rst
```diff
@@ -190,7 +191,7 @@ Example for configure search filter
     use Sonata\AdminBundle\Datagrid\DatagridMapper;
     use Sonata\TranslationBundle\Filter\TranslationFieldFilter;
 
-    class FAQCategoryAdmin extends AbstractAdmin
+    final class FaqCategoryAdmin extends AbstractAdmin
```
kunicmarko20: why change it to `Faq`?

OskarStark: because this is just a docs example and this is valid CamelCase, or am I missing sth? 🤔

OskarStark: friendly ping @kunicmarko20

kunicmarko20: Based on [PSR1](https://www.php-fig.org/psr/psr-1/):

> Class names MUST be declared in StudlyCaps

If I am correct that means *any* letter can be a capital one, which means `FAQ` and both `Faq` are correct, so that is why I asked why the change.

OskarStark: ok, nvm, I revert it 😄

PR #281, opened 2019-02-16
Display locale names instead of flags in language switcher

### LINKED ISSUE #133: flags: language codes vs country codes
In #132 the icon for japanese language was added - we only had jp.png which is the country japan, not the language japanese.

A "locale" can be two letters, representing a language (en, fr, ja, ...). It can also be xx_XX, where the second part in capitals is the country. fr_CA, fr_CH, de_DE, de_CH, pt_BR and so on. In #132 we say that atm, this bundle would use the country code to locate a flag, which is wrong. fr_CA and en_CA is not the same at all, neither are de_CH and fr_CH.

I have seen people using diagonal mixes of the flag of the country that "owns" that language (UK for english, germany for german, france for french, portugal for portuguese, ...) and the other half the flag of the country specified in the locale. I can't find a good example, but a bit like this: http://www.shutterstock.com/pic-98969702/stock-vector-american-and-british-english-language-icon-useful-for-websites-isolated-vector-illustration.html

Technically, i think the sane solution is to allow to provide icons like fr_CA.png and write a twig helper that figures out what image file to include, falling back from fr_CA to fr.

sonofliberty:
## Display locale name instead of flags in language switcher

This pull request removes the country flags from the language switcher and replaces them with the (translated) locale names. 

There is no one-to-one relation between a locale and a country. A language can be spoken in multiple countries and one country can speak multiple languages.

I am targeting 2.x because this change is backwards compatible.

Closes #133

## Changelog

```markdown
### Added
- Feature flag to either keep using country flags or show locale names in language switcher

### Changed
- Replace country flags with locale names in language switcher
```

### DIFF
--- a/src/Block/LocaleSwitcherBlockService.php
+++ b/src/Block/LocaleSwitcherBlockService.php
@@ -31,11 +31,10 @@ class LocaleSwitcherBlockService extends AbstractBlockService
     private $showCountryFlags;
 
     public function __construct(
-        ?string $name = null, 
-        EngineInterface $templating = null, 
+        ?string $name = null,
+        EngineInterface $templating = null,
         ?bool $showCountryFlags = false
-    )
-    {
+    ) {
         parent::__construct($name, $templating);
         $this->showCountryFlags = $showCountryFlags;
     }
--- a/src/Block/LocaleSwitcherBlockService.php
+++ b/src/Block/LocaleSwitcherBlockService.php
@@ -30,7 +30,11 @@ class LocaleSwitcherBlockService extends AbstractBlockService
      */
     private $showCountryFlags;
 
-    public function __construct(?string $name = null, EngineInterface $templating = null, ?bool $showCountryFlags = false)
+    public function __construct(
+        ?string $name = null, 
+        EngineInterface $templating = null, 
+        ?bool $showCountryFlags = false
+    )
     {
         parent::__construct($name, $templating);
         $this->showCountryFlags = $showCountryFlags;
--- a/src/Block/LocaleSwitcherBlockService.php
+++ b/src/Block/LocaleSwitcherBlockService.php
@@ -28,12 +28,12 @@ class LocaleSwitcherBlockService extends AbstractBlockService
     /**
      * @var bool
      */
-    private $useFlags;
+    private $showCountryFlags;
 
-    public function __construct(bool $useFlags, ?string $name = null, EngineInterface $templating = null)
+    public function __construct(?string $name = null, EngineInterface $templating = null, ?bool $showCountryFlags = false)
     {
-        $this->useFlags = $useFlags;
         parent::__construct($name, $templating);
+        $this->showCountryFlags = $showCountryFlags;
     }
 
     /**
@@ -55,7 +55,7 @@ class LocaleSwitcherBlockService extends AbstractBlockService
                 'template' => '@SonataTranslation/Block/block_locale_switcher.html.twig',
                 'locale_switcher_route' => null,
                 'locale_switcher_route_parameters' => [],
-                'locale_switcher_use_flags' => $this->useFlags,
+                'locale_switcher_show_country_flags' => $this->showCountryFlags,
             ]
         );
     }
--- a/src/Block/LocaleSwitcherBlockService.php
+++ b/src/Block/LocaleSwitcherBlockService.php
@@ -15,6 +15,7 @@ namespace Sonata\TranslationBundle\Block;
 
 use Sonata\BlockBundle\Block\BlockContextInterface;
 use Sonata\BlockBundle\Block\Service\AbstractBlockService;
+use Symfony\Bundle\FrameworkBundle\Templating\EngineInterface;
 use Symfony\Component\HttpFoundation\Response;
 use Symfony\Component\OptionsResolver\OptionsResolver;
 use Symfony\Component\OptionsResolver\OptionsResolverInterface;
@@ -25,6 +26,17 @@ use Symfony\Component\OptionsResolver\OptionsResolverInterface;
 class LocaleSwitcherBlockService extends AbstractBlockService
 {
     /**
+     * @var bool
+     */
+    private $useFlags;
+
+    public function __construct(bool $useFlags, ?string $name = null, EngineInterface $templating = null)
+    {
+        $this->useFlags = $useFlags;
+        parent::__construct($name, $templating);
+    }
+
+    /**
      * NEXT_MAJOR: remove this method.
      *
      * @deprecated since 3.x, will be removed in 4.0
@@ -34,9 +46,6 @@ class LocaleSwitcherBlockService extends AbstractBlockService
         $this->configureSettings($resolver);
     }
 
-    /**
-     * {@inheritdoc}
-     */
     public function configureSettings(OptionsResolver $resolver)
     {
         $resolver->setDefaults(
@@ -46,13 +55,11 @@ class LocaleSwitcherBlockService extends AbstractBlockService
                 'template' => '@SonataTranslation/Block/block_locale_switcher.html.twig',
                 'locale_switcher_route' => null,
                 'locale_switcher_route_parameters' => [],
+                'locale_switcher_use_flags' => $this->useFlags,
             ]
         );
     }
 
-    /**
-     * {@inheritdoc}
-     */
     public function execute(BlockContextInterface $blockContext, Response $response = null)
     {
         return $this->renderPrivateResponse($blockContext->getTemplate(), [
--- a/src/DependencyInjection/Configuration.php
+++ b/src/DependencyInjection/Configuration.php
@@ -89,6 +89,11 @@ class Configuration implements ConfigurationInterface
                     ->info('Enable the global locale switcher services.')
                     ->defaultFalse()
                 ->end()
+                // NEXT_MAJOR: Switch default value or remove the flags entirely if mapping of locales to countries has not been fixed.
+                ->booleanNode('locale_switcher_use_flags')
+                    ->info('Whether the language switcher should show languages as flags')
+                    ->defaultTrue()
+                ->end()
             ->end();
 
         return $treeBuilder;
--- a/src/DependencyInjection/Configuration.php
+++ b/src/DependencyInjection/Configuration.php
@@ -89,11 +89,22 @@ class Configuration implements ConfigurationInterface
                     ->info('Enable the global locale switcher services.')
                     ->defaultFalse()
                 ->end()
-                // NEXT_MAJOR: Switch default value or remove the flags entirely if mapping of locales to countries has not been fixed.
-                ->booleanNode('locale_switcher_use_flags')
+                // NEXT_MAJOR: Fix locale to country flag mapping OR remove country flags entirely
+                ->booleanNode('locale_switcher_show_country_flags')
                     ->info('Whether the language switcher should show languages as flags')
                     ->defaultTrue()
                 ->end()
+            ->end()
+            ->beforeNormalization()
+                ->ifTrue(function($v) {return !isset($v['locale_switcher_show_country_flags']) || true === $v['locale_switcher_show_country_flags'];})
+                ->then(function($v) {
+                    @trigger_error(sprintf(
+                        'Showing the country flags is deprecated. The flags will be removed in the next major version unless the maintainers of this bundle fix the locale to country mapping. Please set "%s" to false to avoid this message.',
+                        'sonata_translation.locale_switcher_show_country_flags'
+                    ), E_USER_DEPRECATED);
+                    $v['locale_switcher_show_country_flags'] = isset($v['locale_switcher_show_country_flags']) ? $v['locale_switcher_show_country_flags'] : true;
+                    return $v;
+                })
             ->end();
 
         return $treeBuilder;
--- a/src/DependencyInjection/Configuration.php
+++ b/src/DependencyInjection/Configuration.php
@@ -97,7 +97,7 @@ class Configuration implements ConfigurationInterface
             ->end()
             ->beforeNormalization()
                 ->ifTrue(function ($v) {return !isset($v['locale_switcher_show_country_flags']) || true === $v['locale_switcher_show_country_flags']; })
-                ->then(function($v) {
+                ->then(function ($v) {
                     @trigger_error(sprintf(
                         'Showing the country flags is deprecated. The flags will be removed in the next major version unless the maintainers of this bundle fix the locale to country mapping. Please set "%s" to false to avoid this message.',
                         'sonata_translation.locale_switcher_show_country_flags'
--- a/src/DependencyInjection/Configuration.php
+++ b/src/DependencyInjection/Configuration.php
@@ -96,13 +96,13 @@ class Configuration implements ConfigurationInterface
                 ->end()
             ->end()
             ->beforeNormalization()
-                ->ifTrue(function($v) {return !isset($v['locale_switcher_show_country_flags']) || true === $v['locale_switcher_show_country_flags'];})
+                ->ifTrue(function ($v) {return !isset($v['locale_switcher_show_country_flags']) || true === $v['locale_switcher_show_country_flags']; })
                 ->then(function($v) {
                     @trigger_error(sprintf(
                         'Showing the country flags is deprecated. The flags will be removed in the next major version unless the maintainers of this bundle fix the locale to country mapping. Please set "%s" to false to avoid this message.',
                         'sonata_translation.locale_switcher_show_country_flags'
                     ), E_USER_DEPRECATED);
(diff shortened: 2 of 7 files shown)


src/Twig/Extension/SonataTranslationExtension.php
```diff
@@ -26,6 +28,12 @@ class SonataTranslationExtension extends \Twig_Extension
     protected $translatableChecker;
 
     /**
+     * @var LocaleBundleInterface
+     */
+    protected $localeBundle;
```
greg0ire: Using `protected` everywhere was a mistake on our part. We are trying to move to `private` as much as we can for new fields:

```suggestion
    private $localeBundle;
```

src/Twig/Extension/SonataTranslationExtension.php
```diff
@@ -26,6 +28,12 @@ class SonataTranslationExtension extends \Twig_Extension
...
+
+    /**
+     * SonataTranslationExtension constructor.
```
greg0ire: Please remove this. We know what this is.

```suggestion
```

src/Twig/Extension/SonataTranslationExtension.php
```diff
@@ -57,13 +65,31 @@ public function getTranslatableChecker()
         return $this->translatableChecker;
     }
 
+    /**
+     * @return LocaleBundleInterface
+     */
```
greg0ire: Please remove this comment.

src/Twig/Extension/SonataTranslationExtension.php
```diff
@@ -78,4 +104,14 @@ public function isTranslatable($object)
     {
         return $this->getTranslatableChecker()->isTranslatable($object);
     }
+
+    /**
+     * @param string $locale
+     * @param string|null $displayLocale
+     * @return string|null
+     */
```
greg0ire: Please remove this comment

src/Twig/Extension/SonataTranslationExtension.php
```diff
@@ -78,4 +104,14 @@ public function isTranslatable($object)
...
+    public function getLocaleName(string $locale, ?string $displayLocale = null)
```
greg0ire:
```suggestion
    public function getLocaleName(string $locale, ?string $displayLocale = null): ?string
```

greg0ire:
The build fails for an unrelated issue, I am going to try to fix it in a separate PR.

src/Resources/views/Block/block_locale_switcher.html.twig
```diff
@@ -23,13 +23,13 @@
...
-                    ) }}"
+                    locale_switcher_route,
+                    {'id': admin.id(object), 'tl': locale}|merge(locale_switcher_route_parameters)
+                ) }}"
                    accesskey=""
-                   class=" {% if currentLocale == locale %} active {% endif %}"
+                   class="{% if currentLocale == locale %}active{% endif %}"
                    title="{{ 'admin.locale_switcher.tooltip' |trans([], 'SonataTranslationBundle') }}">
-                    <img src="{{ asset('bundles/sonatatranslation/img/flags/' ~ locale[-2:2]|lower ~ '.png')}}" />
+                    {{ locale|localeName(app.request.locale) }}
```
greg0ire: For now, we should probably leave this files here in case someone overrode that template and is relying on them, but we should remember to remove them later.
```suggestion
                    {# NEXT_MAJOR: remove flag icons #}
                    {{ locale|localeName(app.request.locale) }}
```

greg0ire:
Blocked by #282

greg0ire:
I misclicked, sorry :sweat_smile:

sonofliberty:
Thanks, I updated the pull request. :-)

OskarStark:
Please rebase first, afterwards fix your code style issues please

greg0ire:
I just did it for him :)

src/Twig/Extension/SonataTranslationExtension.php
```diff
@@ -57,6 +64,11 @@ public function getTranslatableChecker()
         return $this->translatableChecker;
     }
 
+    public function getLocaleBundle(): LocaleBundleInterface
+    {
+        return $this->localeBundle instanceof LocaleBundleInterface ? $this->localeBundle : ($this->localeBundle = Intl::getLocaleBundle());
```
kunicmarko20: can't this be done in ctor instead of executing this every time?

src/Twig/Extension/SonataTranslationExtension.php
```diff
@@ -67,6 +79,16 @@ public function getTests()
         ];
     }
 
+    /**
+     * {@inheritdoc}
```
kunicmarko20: please remove this comment

src/Twig/Extension/SonataTranslationExtension.php
```diff
@@ -67,6 +79,16 @@ public function getTests()
...
+     */
+    public function getFilters()
+    {
+        return [
+            new \Twig_SimpleFilter('localeName', [$this, 'getLocaleName']),
```
kunicmarko20: should we use the new namespace?

src/Twig/Extension/SonataTranslationExtension.php
```diff
@@ -78,4 +100,9 @@ public function isTranslatable($object)
     {
         return $this->getTranslatableChecker()->isTranslatable($object);
     }
+
+    public function getLocaleName(string $locale, ?string $displayLocale = null): ?string
```
kunicmarko20: `displayLocale` sounds like a boolean to me, what does this represent?

sonofliberty: It's the language to which the locale is being translated to.

src/Twig/Extension/SonataTranslationExtension.php
sonofliberty: Done

src/Twig/Extension/SonataTranslationExtension.php
sonofliberty: Done

src/Twig/Extension/SonataTranslationExtension.php
sonofliberty: Done

dbu:
makes sense and i can totally confirm, living in switzerland :-D

some people (and their screen designers) might be upset by suddenly not having flags anymore, though. its a visual BC break imho.

i wonder if we should make it a configuration flag whether to use flags or text. then we could do this without any BC break now, and change the default in the next major.

src/Twig/Extension/SonataTranslationExtension.php
```diff
@@ -14,56 +14,59 @@
...
     /**
-     * @param TranslatableChecker $translatableChecker
+     * @var LocaleBundleInterface
      */
+    private $localeBundle;
+
     public function __construct(TranslatableChecker $translatableChecker)
     {
         $this->translatableChecker = $translatableChecker;
+        $this->localeBundle = Intl::getLocaleBundle();
```
OskarStark: If we don’t inject this, why define it in the controller? It can be used directl in the method below or am I missing something?

I am currently on a phone

src/Twig/Extension/SonataTranslationExtension.php
kunicmarko20: What about `selectedLocale`?

sonofliberty:
@dbu I thought about implementing a config flag for this. I didn't do it because the current solution is fundamentally broken imho.

Example: I work with Vietnamese translations and the locale code for Vietnamese is `vi`. But the Vietnamese flag file is named `vn.png` (after the country code). This results in the flag of the Virgin Islands (`vi`) being shown instead of the Vietnamese one. The current solution only works accidentally and for the wrong reasons.

src/Twig/Extension/SonataTranslationExtension.php
sonofliberty: Good point, will take a look.

dbu:
> @dbu I thought about implementing a config flag for this. I didn't do it because the current solution is fundamentally broken imho.
> Example: vi vs vn

yeah, that is confusing and bad. but i would guess quite some people are using this for en / fr / de / it / es in europe, where it happens to look correct. think a minor version should not bring an unexpected visual change. those that do non-european languages where things don't accidentally work will be aware of the problem and be happy to change the configuration. those who are on en / de or en / fr might be quite upset when a minor upgrade breaks their design (in a not obvious way).

my suggestion is to keep the buggy code (with a comment that its incorrect) and a feature flag for it. then the next major can switch the default or remove the buggy code if nobody cared for fixing the implementation to translate language code to country code.

src/Twig/Extension/SonataTranslationExtension.php
sonofliberty: Tbh I just used the same parameter names as defined in [LocaleBundleInterface](https://github.com/symfony/intl/blob/master/ResourceBundle/LocaleBundleInterface.php). As `displayLocale` defines the language in which the locale will be displayed I think the name is appropriate. What do you think?

kunicmarko20: cc @sonata-project/contributors wdyt?

src/DependencyInjection/Configuration.php
```diff
@@ -89,6 +89,11 @@ public function getConfigTreeBuilder()
                     ->info('Enable the global locale switcher services.')
                     ->defaultFalse()
                 ->end()
+                // NEXT_MAJOR: Switch default value or remove the flags entirely if mapping of locales to countries has not been fixed.
```
kunicmarko20: I would remove it in next major since it is wrong and maybe add deprecation if the value is true in the extension, what do others think? cc @sonata-project/contributors

src/Block/LocaleSwitcherBlockService.php
```diff
@@ -24,6 +25,17 @@
  */
 class LocaleSwitcherBlockService extends AbstractBlockService
 {
+    /**
+     * @var bool
+     */
+    private $useFlags;
```
dbu: i would rename this flag to "showCountryFlags" and "locale_switcher_show_country_flags" to be more clear what this is. (as its a flag about flags, which could be confusing ;-) )

dbu:
and we should close #133 with this, and create a new issue to implement a configurable language to flag mapper (#133 is quite a lengthy discussion by now). then if somebody needs images, they hopefully make a pull request for it.

src/Twig/Extension/SonataTranslationExtension.php
greg0ire: Let's be consistent and go with `displayLocale` I suppose

src/DependencyInjection/Configuration.php
greg0ire: I would remove it too

src/Block/LocaleSwitcherBlockService.php
sonofliberty: Done!

src/DependencyInjection/Configuration.php
```diff
@@ -89,11 +89,22 @@ public function getConfigTreeBuilder()
...
                 ->end()
+            ->end()
+            ->beforeNormalization()
+                ->ifTrue(function($v) {return !isset($v['locale_switcher_show_country_flags']) || true === $v['locale_switcher_show_country_flags'];})
+                ->then(function($v) {
+                    @trigger_error(sprintf(
+                        'Showing the country flags is deprecated. The flags will be removed in the next major version unless the maintainers of this bundle fix the locale to country mapping. Please set "%s" to false to avoid this message.',
+                        'sonata_translation.locale_switcher_show_country_flags'
+                    ), E_USER_DEPRECATED);
+                    $v['locale_switcher_show_country_flags'] = isset($v['locale_switcher_show_country_flags']) ? $v['locale_switcher_show_country_flags'] : true;
```
OskarStark: You can use Elvis Operator here

src/Resources/views/Block/block_locale_switcher.html.twig
```diff
@@ -23,13 +24,20 @@
...
                 <a href="{{ admin.generateUrl(
-                        locale_switcher_route,
-                        {'id': admin.id(object), 'tl': locale}|merge(locale_switcher_route_parameters)
-                    ) }}"
+                    locale_switcher_route,
+                    {'id': admin.id(object), 'tl': locale}|merge(locale_switcher_route_parameters)
+                ) }}"
                    accesskey=""
-                   class=" {% if currentLocale == locale %} active {% endif %}"
+                   class="{% if currentLocale == locale %}active{% endif %}"
```
OskarStark: If current != locale do we need the class=„“ afterwards?

I am currently on a phone 📱

src/Twig/Extension/SonataTranslationExtension.php
```diff
@@ -14,56 +14,52 @@
...
     public function __construct(TranslatableChecker $translatableChecker)
     {
         $this->translatableChecker = $translatableChecker;
     }
 
-    /**
-     * {@inheritdoc}
-     */
-    public function getName()
+    public function getName(): string
```
OskarStark: Someone could extend this extension and in this case it would be a be break but I think we can do it in this case

src/DependencyInjection/Configuration.php
sonofliberty: Done!

src/Resources/views/Block/block_locale_switcher.html.twig
sonofliberty: Done!

src/Block/LocaleSwitcherBlockService.php
```diff
@@ -24,6 +25,17 @@
  */
 class LocaleSwitcherBlockService extends AbstractBlockService
 {
+    /**
+     * @var bool
+     */
+    private $showCountryFlags;
+
+    public function __construct(?string $name = null, EngineInterface $templating = null, ?bool $showCountryFlags = false)
```
greg0ire:
```suggestion
    public function __construct(
        ?string $name = null, 
        EngineInterface $templating = null, 
        ?bool $showCountryFlags = false
    )
```

src/DependencyInjection/Configuration.php
```diff
@@ -89,6 +89,23 @@ public function getConfigTreeBuilder()
...
+                ->booleanNode('locale_switcher_show_country_flags')
+                    ->info('Whether the language switcher should show languages as flags')
+                    ->defaultTrue()
+                ->end()
+            ->end()
+            ->beforeNormalization()
+                ->ifTrue(function ($v) {return !isset($v['locale_switcher_show_country_flags']) || true === $v['locale_switcher_show_country_flags']; })
+                ->then(function ($v) {
+                    @trigger_error(sprintf(
+                        'Showing the country flags is deprecated. The flags will be removed in the next major version unless the maintainers of this bundle fix the locale to country mapping. Please set "%s" to false to avoid this message.',
```
greg0ire: I don't know about you @dbu but I don't plan to maintain a long list of hacks.
```suggestion
                        'Showing the country flags is deprecated. The flags will be removed in the next major version. Please set "%s" to false to avoid this message.',
```

src/DependencyInjection/Configuration.php
```diff
@@ -89,6 +89,23 @@ public function getConfigTreeBuilder()
...
                     ->defaultFalse()
                 ->end()
+                // NEXT_MAJOR: Fix locale to country flag mapping OR remove country flags entirely
+                ->booleanNode('locale_switcher_show_country_flags')
+                    ->info('Whether the language switcher should show languages as flags')
+                    ->defaultTrue()
+                ->end()
+            ->end()
+            ->beforeNormalization()
+                ->ifTrue(function ($v) {return !isset($v['locale_switcher_show_country_flags']) || true === $v['locale_switcher_show_country_flags']; })
```
dbu: isn't the value always set when we defaultTrue?

src/Twig/Extension/SonataTranslationExtension.php
dbu: i'd ignore that. and in the next major we should make all classes final unless they are actively intended to be extended to avoid such worries in the future.

src/DependencyInjection/Configuration.php
sonofliberty: Actually no. At this point `$v` only contains whatever you have in your configuration.

dbu: ah right, beforeNormalization :man_facepalming:

sonofliberty:
Sorry I'm kinda new to this process. When could I expect this PR to be merged? Or is there anything left to do?

Cheers

dbu:
i am not one of the maintainers of the library, so i don't feel authorized to press the merge button.

what should be added (here or by one of the maintainers) is a changelog entry (with a new heading 2.5.0 - unreleased) that mentions the new config option.

OskarStark:
Thank you @sonofliberty 🙏 I squashed your commits into one!

@dbu we will handle the change log on release, based on the PR header part to avoid too many merge conflicts for other PR’s 😉

PR #286, opened 2019-03-25
Added string typecheck to prevent exception

jeroendk:
## Added String typecheck to the AdminExtensionCompilerPass file before calling class_exists()

When defining an Admin service the Entity argument can be left empty. This causes an Exception in the `AdminExtensionCompilerPass` when calling class_exists(), because strict_types is set to 1.  

I am targeting this branch, because this bugfix is backwards compatible. 

## Changelog
```markdown
### Fixed
Added is_string check in AdminExtensionCompilerPass
```


jeroendk:
Sure! Give some time to do so.

OskarStark:
Any news on this PR?

PR #287, opened 2019-04-01
Make it possible to create an admin without entity

quisse:
Currently when creating an admin with no entity (`~` as argument), the [get_class](https://github.com/storefront-bvba/SonataTranslationBundle/blob/af09eb0f51162ae05c5365835869f4ddfeed9a98/src/DependencyInjection/Compiler/AdminExtensionCompilerPass.php#L36) throws `class_exists() expects parameter 1 to be string`. To overcome this I would first check if there's a model class and after that check if it exists. 
An other solution is to provide an empty string as argument but that's a bit hacky.

## Subject

I am targeting this branch, because it's a small improvement which AFAIK wont break a thing.

## Changelog

```markdown
### Fixed
- Admin without entity shouldn't be restricted
```


greg0ire:
Thanks @quisse !

PR #304, opened 2019-10-02
Fixed issue caused by always using the default entity manager

MitchelAnthony:
## Fixed issue caused by always using the default entity manager

When using multiple entity managers with different entities in their chains, this extension always fetched the default entity manager from the registry instead of the one for the given object

I am targeting this branch, because it is a BC bug fix.

## Changelog

```markdown
### Fixed
Fixed issue caused by always using the default entity manager
```


phansys:
Thank you so much for the fix!
Please, apply the patch provided by [flintci](https://flintci.io/repositories/89/analyses/24624).

OskarStark:
No need for bothering, FlintCI will fix it afterwards

PR #332, opened 2020-03-27
Drop SonataCoreBundle

wbloszyk:
## Subject

I am targeting this branch, because it its BC-~~break~~.

## Changelog

```markdown
### Removed
- Removed `SonataCoreBundle` 
```

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -23,7 +23,6 @@
         "php": "^7.1",
         "sonata-project/admin-bundle": "^3.43",
         "sonata-project/block-bundle": "^3.11",
-        "sonata-project/core-bundle": "^3.12.0",
         "symfony/config": "^3.4 || ^4.2",
         "symfony/dependency-injection": "^3.4 || ^4.2",
         "symfony/form": "^3.4 || ^4.2",


VincentLanglet:
> I am targeting this branch, because it its BC-break.

Why ?

wbloszyk:
> > I am targeting this branch, because it its BC-break.
> 
> Why ?

Becouse It replace SonataCoreBundle to extensions packkages. They are in conflict, so You can not use both. User must upgrade bundles (like admin) by major version.

VincentLanglet:
> > > I am targeting this branch, because it its BC-break.
> > 
> > 
> > Why ?
> 
> Becouse It replace SonataCoreBundle to extensions packages. They are in conflict, so You can not use both. User must upgrade bundles (like admin) by major version.

IMO, if you're just dropping a dependency, this is not a BC-break, since composer will resolve the right version to use for the user depending on the constraint.
But maybe I'm wrong.

phansys:
If none of the features provided by the removed package are used in this package, there is no BC break at all.

wbloszyk:
My bad, you have right. I change branch from master to 2.x.

phansys:
[Travis is passing](https://travis-ci.org/github/sonata-project/SonataTranslationBundle/builds/667596315?), but no reporting (yet).

greg0ire:
Thanks @wbloszyk !

Issue #335, opened 2020-04-02
twig/intl-extra is required since 2.5.0


jorrit:
#### Sonata packages

```
$ composer show --latest 'sonata-project/*'
sonata-project/admin-bundle              3.64.0 3.64.0 The missing Symfony Admin Generator
sonata-project/block-bundle              3.18.4 4.2.0  Symfony SonataBlockBundle
sonata-project/cache                     2.0.1  2.0.1  Cache library
sonata-project/core-bundle               3.18.0 3.18.0 Symfony SonataCoreBundle (abandoned)
sonata-project/datagrid-bundle           2.5.0  3.1.1  Symfony SonataDatagridBundle
sonata-project/doctrine-extensions       1.6.0  1.6.0  Doctrine2 behavioral extensions
sonata-project/doctrine-orm-admin-bundle 3.15.0 3.15.0 Symfony Sonata / Integrate Doctrine ORM into the SonataAdminBundle
sonata-project/exporter                  2.2.0  2.2.0  Lightweight Exporter library
sonata-project/intl-bundle               2.7.0  2.7.0  Symfony SonataIntlBundle
sonata-project/translation-bundle        2.5.0  2.5.0  SonataTranslationBundle
```

#### Symfony packages

```
$ composer show --latest 'symfony/*'

symfony/asset                      v4.4.7  v4.4.7  Symfony Asset Component
symfony/browser-kit                v4.4.7  v4.4.7  Symfony BrowserKit Component
symfony/cache                      v4.4.7  v4.4.7  Symfony Cache component with PSR-6, PSR-16, and tags
symfony/cache-contracts            v2.0.1  v2.0.1  Generic abstractions related to caching
symfony/config                     v4.4.7  v4.4.7  Symfony Config Component
symfony/console                    v4.4.7  v4.4.7  Symfony Console Component
symfony/css-selector               v4.4.7  v4.4.7  Symfony CssSelector Component
symfony/debug                      v4.4.7  v4.4.7  Symfony Debug Component
symfony/debug-bundle               v4.4.7  v4.4.7  Symfony DebugBundle
symfony/debug-pack                 v1.0.7  v1.0.7  A debug pack for Symfony projects
symfony/dependency-injection       v4.4.7  v4.4.7  Symfony DependencyInjection Component
symfony/doctrine-bridge            v4.4.7  v4.4.7  Symfony Doctrine Bridge
[...]

PR #336, opened 2020-04-02
Drop support of Symfony < 4.3

franmomu:
## Subject

Before solving https://github.com/sonata-project/SonataTranslationBundle/issues/335, dropping Symfony < 4.3 and some tests clean up 

I am targeting this branch, because this is BC.

## Changelog

```markdown
### Removed
- Support of Symfony < 4.3
```

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -21,17 +21,18 @@
     ],
     "require": {
         "php": "^7.1",
-        "sonata-project/admin-bundle": "^3.43",
-        "sonata-project/block-bundle": "^3.11",
-        "symfony/config": "^3.4 || ^4.2",
-        "symfony/dependency-injection": "^3.4 || ^4.2",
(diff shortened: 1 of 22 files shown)


src/EventSubscriber/LocaleSubscriber.php
```diff
@@ -35,7 +35,7 @@ public function __construct($defaultLocale = 'en')
         $this->defaultLocale = $defaultLocale;
     }
 
-    public function onKernelRequest(GetResponseEvent $event)
+    public function onKernelRequest(RequestEvent $event)
```
phansys, line 38: Isn't this a BC break?

franmomu: Yes, thanks, I saw it was final 🤦‍♂and didn't think too much.

PR #345, opened 2020-05-28
Replace usage of deprecated spaceless Twig tag with apply spaceless

jorrit:
## Subject

This PR fixes the following deprecation notice:

> The spaceless tag in "@SonataTranslation/Block/block_locale_switcher.html.twig" at line 17 is deprecated since Twig 2.7, use the "spaceless" filter with the "apply" tag instead.

I am targeting this branch, because Twig 2.12 is the minimum required version of this branch.

## Changelog

```markdown
### Fixed
- Deprecation notice caused by use of spaceless tag in block_locale_switcher.html.twig.
```

### DIFF
--- a/src/Resources/views/Block/block_locale_switcher.html.twig
+++ b/src/Resources/views/Block/block_locale_switcher.html.twig
@@ -14,7 +14,7 @@
     {% endfor %}
 
     <div class="locale_switcher">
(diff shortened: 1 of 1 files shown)


phansys:
Thank you @jorrit!

PR #355, opened 2020-08-25
[FEAT] : Add new features to LocaleSwitcherBlockService

c-Rolland:
## Subject

I am targeting this branch to add the following features to **LocaleSwitcherBlockService** : <br />

### To **hidden** or **display** the locale switcher block.
Capture d’écran 2020-08-25 à 21 19 35

```
sonata_translation:
        display_locale_switcher_block: false // the default is true
```

### To override the block template :
```
sonata_translation:
        locale_switcher_block_template: '@App/Block/custom_locale_switcher.html.twig'

```

### Wrap the block template with 2 blocks to override it easier if needed.

```
{% block main_translation_locale_switcher_block %} {% endblock %}
{% block translation_locale_switcher %} {% endblock %}
```


src/Block/LocaleSwitcherBlockService.php
```diff
@@ -61,10 +76,11 @@ public function configureSettings(OptionsResolver $resolver)
             [
                 'admin' => null,
                 'object' => null,
-                'template' => '@SonataTranslation/Block/block_locale_switcher.html.twig',
+                'template' => $this->template ?? '@SonataTranslation/Block/block_locale_switcher.html.twig',
```
jordisala1991, line 79: Isnt it better to just override the template at twig level? I mean creating the same file on your templates directory: https://symfony.com/doc/current/bundles/override.html#templates

src/Resources/views/Block/block_locale_switcher.html.twig
```diff
@@ -1,39 +1,52 @@
...
-{% endif %}
+        {% for extension in admin.extensions %}
+            {% if extension.translatableLocale is defined %}
+                {% set currentLocale = extension.translatableLocale(admin) %}
+            {% endif %}
+        {% endfor %}
+
+        {% block translation_locale_switcher %}
+            <div class="locale_switcher">
+                {% spaceless %}
```
phansys, line 22: `{% spaceless %}` tag is deprecated, please use `{% apply spaceless %}` instead.

PR #369, opened 2020-09-03
Update `UserLocaleSubscriber::onInteractiveLogin()` in order to avoid starting the session

phansys:
## Subject
Update `UserLocaleSubscriber::onInteractiveLogin()` in order to avoid starting the session.

I am targeting this branch, because this change respects BC.

## Changelog

```markdown
### Fixed
- Fixed starting the session in `UserLocaleSubscriber::onInteractiveLogin()` when there is no previous session.
```

### DIFF
--- a/src/EventSubscriber/UserLocaleSubscriber.php
+++ b/src/EventSubscriber/UserLocaleSubscriber.php
@@ -14,7 +14,6 @@ declare(strict_types=1);
 namespace Sonata\TranslationBundle\EventSubscriber;
 
 use Symfony\Component\EventDispatcher\EventSubscriberInterface;
-use Symfony\Component\HttpFoundation\Session\SessionInterface;
 use Symfony\Component\Security\Http\Event\InteractiveLoginEvent;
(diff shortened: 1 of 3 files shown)


n4huel (approved):

src/EventSubscriber/UserLocaleSubscriber.php
```diff
@@ -38,6 +38,10 @@ public function __construct(SessionInterface $session)
 
     public function onInteractiveLogin(InteractiveLoginEvent $event)
     {
+        if (!$this->session->isStarted()) {
```
franmomu, line 41:
```suggestion
        if (!$event->getRequest()->hasPreviousSession()) {
```

Since the `InteractiveLoginEvent` has the `request` object, what about removing the `$session` attribute of this class and retrieving the session from the request? I don't remember exactly, but sometimes simple accessing the session, starts it.

phansys: Thank you @franmomu. Your suggestion is addressed.

franmomu (approved):

PR #415, opened 2020-11-06
Psalm phpstan levels

franmomu:
This raises from 1 to 6 the phpstan level and from 6 to 4 the psalm one.

### DIFF
--- a/phpstan.neon.dist
+++ b/phpstan.neon.dist
@@ -1,8 +1,9 @@
 includes:
     - phpstan-baseline.neon
+    - vendor/phpstan/phpstan-symfony/extension.neon
 
 parameters:
-    level: 1
+    level: 6
 
     paths:
         - src
--- a/src/Admin/Extension/AbstractTranslatableAdminExtension.php
+++ b/src/Admin/Extension/AbstractTranslatableAdminExtension.php
@@ -84,6 +84,8 @@ abstract class AbstractTranslatableAdminExtension extends AbstractAdminExtension
      * Return current translatable locale
      * ie: the locale used to load object translations != current request locale.
      *
+     * @phpstan-param AdminInterface<object> $admin
+     *
      * @return string
      */
     public function getTranslatableLocale(AdminInterface $admin)
@@ -118,6 +120,8 @@ abstract class AbstractTranslatableAdminExtension extends AbstractAdminExtension
      *
      * @deprecated since version 2.x, to be removed in 3.0. Use dependency injection instead.
      *
+     * @phpstan-param AdminInterface<object> $admin
+     *
      * @return ContainerInterface
      */
     protected function getContainer(AdminInterface $admin)
@@ -128,6 +132,8 @@ abstract class AbstractTranslatableAdminExtension extends AbstractAdminExtension
             __METHOD__
         ), E_USER_DEPRECATED);
 
+        \assert(\is_callable([$admin, 'getConfigurationPool']));
+
         return $admin->getConfigurationPool()->getContainer();
     }
 
@@ -136,9 +142,11 @@ abstract class AbstractTranslatableAdminExtension extends AbstractAdminExtension
      *
      * @deprecated since version 2.x, to be removed in 3.0.
      *
+     * @phpstan-param AdminInterface<object> $admin
+     *
      * Return the list of possible locales for your models.
      *
-     * @return array
+     * @return string[]
      */
     protected function getTranslationLocales(AdminInterface $admin)
     {
@@ -156,6 +164,8 @@ abstract class AbstractTranslatableAdminExtension extends AbstractAdminExtension
      * @deprecated since version 2.x, to be removed in 3.0.
      *
      * Return the default locale if url parameter is not present.
+     *
+     * @phpstan-param AdminInterface<object> $admin
      *
      * @return string
      */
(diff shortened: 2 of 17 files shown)


VincentLanglet (approved):
Why the psalm downgrade ?

franmomu:
> Why the psalm downgrade ?

[In Psalm level 1 is stricter that 8, so the less the better](https://psalm.dev/docs/running_psalm/error_levels/).

phpstan.neon.dist
```diff
@@ -1,11 +1,14 @@
...
+    - vendor/phpstan/phpstan-symfony/extension.neon
 
 parameters:
-    level: 1
+    level: 6
 
     paths:
         - src
 
+    treatPhpDocTypesAsCertain: false
```
greg0ire, line 11: Why is this needed?

src/Admin/Extension/AbstractTranslatableAdminExtension.php
```diff
@@ -84,6 +84,8 @@ public function getTranslatableChecker()
      * Return current translatable locale
      * ie: the locale used to load object translations != current request locale.
      *
+     * @phpstan-param AdminInterface<object> $admin
```
greg0ire, line 87: Should it be
```suggestion
     * @phpstan-param AdminInterface<TranslatableInterface> $admin
```

?

src/DependencyInjection/Compiler/AdminExtensionCompilerPass.php
```diff
@@ -53,6 +56,8 @@ public function process(ContainerBuilder $container)
     }
 
     /**
+     * @param array<int, int|string> $types
```
greg0ire, line 59: The index does not matter, I'm not sure what the best way to express this is…
I'd say either `list<int|string>` to tell people they should not have keys, or `array<int|string>` to tell them they can do whatever they want with the keys.

src/DependencyInjection/SonataTranslationExtension.php
```diff
@@ -102,6 +106,8 @@ public function load(array $configs, ContainerBuilder $container)
     /**
      * @param array $translationTargets
      *
+     * @phpstan-param array<string, array<string, class-string[]>> $translationTargets
```
greg0ire, line 109: That variable is only ever used in a foreach, and the key is not used…
```suggestion
     * @phpstan-param iterable<string, array<string, class-string[]>> $translationTargets
```

or even
```suggestion
     * @phpstan-param iterable<array<string, class-string[]>> $translationTargets
```

phpstan.neon.dist
VincentLanglet: Oh I missed this ! I think we should add this indeed !

greg0ire: Why?

VincentLanglet: Sorry. I meant "shouldnt".

VincentLanglet: We're not supposed to check ourself if the value provided is respecting the phpdoc IMHO.

franmomu: yes, you are right, I added it at the beginning of this PR and then I forgot to remove it and check. The fail was related to the constructor of `LocaleSwitcherBlockService`.

src/Admin/Extension/AbstractTranslatableAdminExtension.php
franmomu: If I do this I get:
```
------ -------------------------------------------------------------------------------------------------------------------------------------------------------
  Line   Admin/Extension/AbstractTranslatableAdminExtension.php
 ------ -------------------------------------------------------------------------------------------------------------------------------------------------------
  109    Parameter #1 $admin of method Sonata\TranslationBundle\Admin\Extension\AbstractTranslatableAdminExtension::getTranslatableLocale() expects
         Sonata\AdminBundle\Admin\AdminInterface<Sonata\TranslationBundle\Model\TranslatableInterface>, Sonata\AdminBundle\Admin\AdminInterface<object> given.
  115    Parameter #1 $admin of method Sonata\TranslationBundle\Admin\Extension\AbstractTranslatableAdminExtension::getTranslatableLocale() expects
         Sonata\AdminBundle\Admin\AdminInterface<Sonata\TranslationBundle\Model\TranslatableInterface>, Sonata\AdminBundle\Admin\AdminInterface<object> given.
 ------ -------------------------------------------------------------------------------------------------------------------------------------------------------
```

In line 109 we have:

https://github.com/sonata-project/SonataTranslationBundle/blob/5a93fc6176739a03f4481930d3cb05a341a2917c/src/Admin/Extension/AbstractTranslatableAdminExtension.php#L106-L109

This method is defined in [`AdminExtensionInterface`](https://github.com/sonata-project/SonataAdminBundle/blob/f2dba48d52ecd60186a7f4f530131f9fd4e1ed62/src/Admin/AdminExtensionInterface.php#L142-L149) as:

```php
/**
 * Get a chance to add persistent parameters.
 *
 * @return array<string, mixed>
 *
 * @phpstan-param AdminInterface<object> $admin
 */
public function getPersistentParameters(AdminInterface $admin);
```

I think the problem seems to be related to https://github.com/sonata-project/SonataAdminBundle/pull/6576, because looks like you cannot substitute `object` by `TranslatableInterface`.

greg0ire (approved):

VincentLanglet (approved):

VincentLanglet:
Thanks !

PR #416, opened 2020-11-06
Allow knplabs/doctrine-behaviors 2

franmomu:
## Subject

Allows to install knplabs/doctrine-behaviors 2, apparently it introduced some interfaces.

## Changelog

```markdown
### Added
- Added support for `knplabs/doctrine-behaviors` 2.
```

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -36,14 +36,14 @@
         "doctrine/persistence": "<1.3.4",
         "doctrine/phpcr-odm": ">=3.0",
         "gedmo/doctrine-extensions": "<2.4.36",
-        "knplabs/doctrine-behaviors": "<1.0 || >=2.0",
+        "knplabs/doctrine-behaviors": "<1.0 || >=3.0",
         "stof/doctrine-extensions-bundle": "<1.1 || >=2.0"
     },
     "require-dev": {
         "doctrine/orm": "^2.5",
         "doctrine/phpcr-odm": "^1.5",
         "jackalope/jackalope-doctrine-dbal": "^1.5",
-        "knplabs/doctrine-behaviors": "^1.4",
+        "knplabs/doctrine-behaviors": "^1.4 || ^2.0",
         "matthiasnoback/symfony-config-test": "^4.2",
         "matthiasnoback/symfony-dependency-injection-test": "^4.0",
         "phpstan/extension-installer": "^1.0",
--- a/tests/Fixtures/Traits/ORM/Knplabs/AbstractArticleTranslation.php
+++ b/tests/Fixtures/Traits/ORM/Knplabs/AbstractArticleTranslation.php
@@ -1,0 +1,46 @@
+<?php
+
+namespace Sonata\TranslationBundle\Tests\Fixtures\Traits\ORM\Knplabs;
+
+use Doctrine\ORM\Mapping as ORM;
+
+abstract class AbstractArticleTranslation
+{
+    /**
+     * @var int|null
+     *
+     * @ORM\Id
+     * @ORM\GeneratedValue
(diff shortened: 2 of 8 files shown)


composer.json
```diff
@@ -36,14 +36,14 @@
         "doctrine/persistence": "<1.3.4",
         "doctrine/phpcr-odm": ">=3.0",
         "gedmo/doctrine-extensions": "<2.4.36",
-        "knplabs/doctrine-behaviors": "<1.0 || >=2.0",
+        "knplabs/doctrine-behaviors": "<1.0 || >=3.0",
```
VincentLanglet, line 39: Why are why adding a conflict with a version 3.0 which doesn't exist currently, but we're not doing the same for `doctrine/annotations` or `doctrine/persistence` or `gedmo/doctrine-extensions` ?

tests/Fixtures/Model/Knplabs/AbstractTranslatableEntity.php
```diff
@@ -0,0 +1,33 @@
+<?php
+
+namespace Sonata\TranslationBundle\Tests\Fixtures\Model\Knplabs;
+
+use Sonata\TranslationBundle\Model\TranslatableInterface;
+
+abstract class AbstractTranslatableEntity implements TranslatableInterface
```
VincentLanglet, line 7: Should it be internal ?

tests/Fixtures/Traits/ORM/Knplabs/AbstractArticleTranslation.php
```diff
@@ -0,0 +1,46 @@
+<?php
+
+namespace Sonata\TranslationBundle\Tests\Fixtures\Traits\ORM\Knplabs;
+
+use Doctrine\ORM\Mapping as ORM;
+
+abstract class AbstractArticleTranslation
```
VincentLanglet, line 7: Should it be internal ?

VincentLanglet (approved):

composer.json
franmomu: Since we are not requiring it in the `require` section, I guess this is to make sure that works in that version. If version 3.0 is released with changes that breaks BC, the bundle will fail.

The other cases are special:
- `doctrine/annotations`: I remember bumping this one because of compatibilities with php 7.4.
- `doctrine/persistence`: This one was to make sure the namespace change.
- `gedmo/doctrine-extensions`: This one was to fix a type declaration.

This bundles supports Gedmo, Knplabs and PHPCR, those libraries do have constraints:

```
"doctrine/phpcr-odm": ">=3.0",
"knplabs/doctrine-behaviors": "<1.0 || >=2.0",
"stof/doctrine-extensions-bundle": "<1.1 || >=2.0"
```

BTW, should we keeping supporting phpcr since our bundle is marked as abandoned in the next major versions?

tests/Fixtures/Model/Knplabs/AbstractTranslatableEntity.php
franmomu: It's in the `tests` directory, I don't think it is needed.

VincentLanglet: Oh, I missed this

composer.json
VincentLanglet: > BTW, should we keeping supporting phpcr since our bundle is marked as abandoned in the next major versions?

If it doesn't add additional work, I would say yes ; maybe we will find a maintainer for the phpcr package one day...

VincentLanglet (approved):

VincentLanglet (approved):

OskarStark (approved):

PR #418, opened 2020-11-08
Raise Psalm to error level 3 and add tests directory

franmomu:
The ignored errors are going to be removed in the next major version.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -46,14 +46,20 @@
         "knplabs/doctrine-behaviors": "^1.4",
         "matthiasnoback/symfony-config-test": "^4.2",
         "matthiasnoback/symfony-dependency-injection-test": "^4.0",
+        "phpstan/extension-installer": "^1.0",
         "phpstan/phpstan": "^0.12.48",
+        "phpstan/phpstan-phpunit": "^0.12.16",
         "phpstan/phpstan-symfony": "^0.12.11",
+        "phpunit/phpunit": "^9.4.0",
+        "psalm/plugin-phpunit": "^0.12",
+        "psalm/plugin-symfony": "^1.5",
         "sonata-project/doctrine-orm-admin-bundle": "^3.2",
         "stof/doctrine-extensions-bundle": "^1.1",
         "symfony/framework-bundle": "^4.4",
         "symfony/phpunit-bridge": "^5.1.8",
         "symfony/templating": "^4.4 || ^5.1",
-        "vimeo/psalm": "^4.1"
+        "vimeo/psalm": "^3.16",
+        "weirdan/doctrine-psalm-plugin": "^0.11.3"
     },
     "suggest": {
         "doctrine/phpcr-odm": "if you translate odm documents",
--- a/phpstan-baseline.neon
+++ b/phpstan-baseline.neon
@@ -1,41 +1,53 @@
 parameters:
-	ignoreErrors:
-		-
-			message: "#^Call to an undefined method Sonata\\\\AdminBundle\\\\Model\\\\ModelManagerInterface\\:\\:getDocumentManager\\(\\)\\.$#"
-			count: 1
-			path: src/Admin/Extension/Phpcr/TranslatableAdminExtension.php
+    ignoreErrors:
+        -
+            # sonata-project/doctrine-phpcr-admin-bundle cannot be currently installed because it is not compatible with sonata-project/block-bundle 4
+            message: "#^Call to an undefined method Sonata\\\\AdminBundle\\\\Model\\\\ModelManagerInterface\\:\\:getDocumentManager\\(\\)\\.$#"
+            count: 1
+            path: src/Admin/Extension/Phpcr/TranslatableAdminExtension.php
 
-		-
-			message: "#^Call to function is_object\\(\\) with string\\|null will always evaluate to false\\.$#"
-			count: 1
-			path: src/Block/LocaleSwitcherBlockService.php
+        # src/Block/LocaleSwitcherBlockService.php issues will be solved in next major version
 
-		-
-			message: "#^Result of \\|\\| is always true\\.$#"
-			count: 1
-			path: src/Block/LocaleSwitcherBlockService.php
+        -
+            message: "#^Call to function is_object\\(\\) with string\\|null will always evaluate to false\\.$#"
+            count: 1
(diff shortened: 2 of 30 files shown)


phpstan-baseline.neon
```diff
@@ -39,3 +39,13 @@ parameters:
...
 			count: 13
 			path: src/Filter/TranslationFieldFilter.php
+
+		-
+			message: "#^Parameter \\#2 \\$showCountryFlagsOrTemplating of class Sonata\\\\TranslationBundle\\\\Block\\\\LocaleSwitcherBlockService constructor expects bool\\|Symfony\\\\Bundle\\\\FrameworkBundle\\\\Templating\\\\EngineInterface\\|null, stdClass given\\.$#"
+			count: 1
+			path: tests/Block/LocaleSwitcherBlockServiceTest.php
+
+		-
+			message: "#^Method Sonata\\\\TranslationBundle\\\\Tests\\\\Fixtures\\\\Model\\\\Knplabs\\\\TranslatableEntity\\:\\:getLocale\\(\\) should return string but returns Knp\\\\DoctrineBehaviors\\\\Model\\\\Translatable\\\\Returns\\.$#"
```
franmomu, line 49: https://github.com/KnpLabs/DoctrineBehaviors/blob/072078422bcada395906ec96b192c6b30a93bebb/src/Model/Translatable/TranslatableMethods.php#L149

The phpdoc is wrong, this will be removed with [knplabs/doctrine-behaviors 2](https://github.com/sonata-project/SonataTranslationBundle/pull/416)

VincentLanglet: Shouldnt we add a comment for every ignored error in order to know when they will be solve ; like we did for SonataAdmin

VincentLanglet (dismissed):

composer.json
```diff
@@ -37,6 +37,7 @@
         "doctrine/phpcr-odm": ">=3.0",
         "gedmo/doctrine-extensions": "<2.4.36",
         "knplabs/doctrine-behaviors": "<1.0 || >=2.0",
+        "phpunit/phpunit": "<8.5.9",
```
VincentLanglet, line 40: We should avoid this conflict.

Isnt it better to update the CI in order to set the version of phpunit to 8.5.9 in the lowest build ?

OskarStark: @jordisala1991 what do you think?

franmomu: Well, the problem here is that `weirdan/doctrine-psalm-plugin` and `psalm/plugin-phpunit` requires `phpunit/phpunit` and we don't require it explicitly (we use `symfony/phpunit-bridge`), this is a specific case, so I don't know what is best.

VincentLanglet: `phpstan/phpstan-phpunit` seems to work only because another package is requiring `phpunit/phpunit` too.

In my personal project, it's not the case, and I had to add in the config.
```
    bootstrapFiles:
        - symfony/bin/.phpunit/phpunit-9.3-0/vendor/autoload.php
```
And to add this in `composer.json` in order to automatically run `bin/phpunit install` 
```
    "scripts": {
        "auto-scripts": {
            ...
        },
        "post-install-cmd": [
            "[ $COMPOSER_DEV_MODE -eq 0 ] || bin/phpunit install",
            "@auto-scripts"
        ],
        "post-update-cmd": [
            "[ $COMPOSER_DEV_MODE -eq 0 ] || bin/phpunit install",
            "@auto-scripts"
        ]
```
But I only use phpstan, not psalm.

IMHO if we need `phpunit/phpunit` for our psalm plugin it's better to add the devRequirement instead of the conflict.

VincentLanglet:
Did you saw my comment: https://github.com/sonata-project/SonataTranslationBundle/pull/418#discussion_r521940210

> IMHO if we need phpunit/phpunit for our psalm plugin it's better to add the devRequirement instead of the conflict.

franmomu:
> Did you saw my comment: [#418 (comment)](https://github.com/sonata-project/SonataTranslationBundle/pull/418#discussion_r521940210)
> 
> > IMHO if we need phpunit/phpunit for our psalm plugin it's better to add the devRequirement instead of the conflict.

Yes I did, but IMHO if we are not requiring it in our code I don't see a problem having a conflict with it and if those packages restrict the phpunit version we don't have to change it. But let's see if someone else has other solution or something.

VincentLanglet:
> Yes I did, but IMHO if we are not requiring it in our code I don't see a problem having a conflict with it.

The issue is the following.

If someone use `Phpunit 8` and `SonataTranslationBundle` on his bundle, he won't be able to update the `SonataTranslationBundle` anymore because of the conflict (which is just our dev-conflict and should not impact someone installing our packaging).

franmomu:
> > Yes I did, but IMHO if we are not requiring it in our code I don't see a problem having a conflict with it.
> 
> The issue is the following.
> 
> If someone use `Phpunit 8` and `SonataTranslationBundle` on his bundle, he won't be able to update the `SonataTranslationBundle` anymore because of the conflict (which is just our dev-conflict and should not impact someone installing our packaging).

Totally right, I don't know why I was thinking about `conflict` being `conflict-dev` (which does not exist, but it may be useful).

VincentLanglet:
> Totally right, I don't know why I was thinking about `conflict` being `conflict-dev` (which does not exist, but it may be useful).

I agree that `conflict-dev` could be useful in some case.

phpstan.neon.dist
```diff
@@ -1,12 +1,14 @@
 includes:
     - phpstan-baseline.neon
+    - vendor/phpstan/phpstan-phpunit/extension.neon
```
VincentLanglet, line 3: As discuss somewhere else, we should use the phpstan installer instead

VincentLanglet (approved):

OskarStark (approved):

OskarStark:
Thank you!

VincentLanglet:
@franmomu Since https://github.com/psalm/psalm-plugin-phpunit/pull/100/files we shouldn't require phpunit now.

franmomu:
> @franmomu Since https://github.com/psalm/psalm-plugin-phpunit/pull/100/files we shouldn't require phpunit now.

cool, if you want to make a PR with what would be the proper way of using it.

VincentLanglet:
I opened https://github.com/sonata-project/SonataTranslationBundle/pull/438 @franmomu

PR #453, opened 2021-01-27
Allow support PHP 8.0

Yozhef:
## Changelog

```markdown
### Added
- Add support for PHP 8.0
```

PR #464, opened 2021-02-25
Remove stof/doctrine-extensions-bundle dependency

franmomu:
## Subject

Ref: https://github.com/sonata-project/SonataTranslationBundle/pull/453#issuecomment-785344620

Instead of relying on the bundle, a `sonata_translation.gedmo.translatable_listener_service` option has been added to the configuration so the user must specify the translatable listener service name when using gedmo (allowing to use another bundle or only the package). 

## Changelog

```markdown
### Added
- Added `sonata_translation.gedmo.translatable_listener_service` option to specify the translatable listener service name.
### Deprecated
- Deprecated not setting `sonata_translation.gedmo.translatable_listener_service` option when using gedmo.
```

### DIFF
--- a/src/DependencyInjection/Configuration.php
+++ b/src/DependencyInjection/Configuration.php
@@ -57,6 +57,23 @@ class Configuration implements ConfigurationInterface
                             ->requiresAtLeastOneElement()
                             ->prototype('scalar')->end()
                         ->end()
+                        ->scalarNode('translatable_listener_service')
+                            ->isRequired()
+                            ->cannotBeEmpty()
+                            ->info('The translatable listener service that has to be registered when using gedmo/doctrine-extensions')
+                            // NEXT_MAJOR: Remove default value.
+                            ->defaultValue(static function (): string {
+                                @trigger_error(
+                                    'Not specifying the translatable listener service when using'
+                                    .' gedmo/doctrine-extensions is deprecated since sonata-project/translation-bundle 2.x.'
+                                    .' If you are using a bundle that integrates this library, it is likely that it'
+                                    .' registers that service for you. In case of "stof/doctrine-extensions-bundle", it is:'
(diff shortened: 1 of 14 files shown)


jordisala1991:
Maybe we can generate the listener ourselves and register it if there is no value provided?

franmomu:
I started doing that, but looking at this service definition:

https://github.com/stof/StofDoctrineExtensionsBundle/blob/f3c948b390f82a357d814331e6ea47bcda813a38/Resources/config/translatable.xml#L12

looks like it's configurable and maybe the user wants to register its own. Anyway if no service name is provided we could register a simple one, I'll change it.

src/DependencyInjection/Configuration.php
```diff
@@ -57,6 +57,23 @@ public function getConfigTreeBuilder()
                             ->requiresAtLeastOneElement()
                             ->prototype('scalar')->end()
                         ->end()
+                        ->scalarNode('translatable_listener_service')
+                            ->isRequired()
+                            ->cannotBeEmpty()
+                            ->info('The translatable listener service that has to be registered when using gedmo/doctrine-extensions')
+                            // NEXT_MAJOR: Remove default value.
+                            ->defaultValue(static function (): string {
```
VincentLanglet, line 65: I don't understand, in the doc we're saying:
```
translatable_listener_service: Gedmo\Translatable\TranslatableListener
```

Shouldn't it be the default value ?

franmomu: It has to be `stof_doctrine_extensions.listener.translatable` to be BC.

jordisala1991 (approved):

VincentLanglet (approved):

jordisala1991:
Thank you @franmomu

PR #465, opened 2021-03-02
2.x into Master

VincentLanglet:


wbloszyk (dismissed):
IMO we should not increase phpstan level in master yet.

VincentLanglet:
> IMO we should not increase phpstan level in master yet.

This was increased in 3.x

wbloszyk:
> > IMO we should not increase phpstan level in master yet.
> 
> This was increased in 3.x

I know. It should be inrease in seperate PR. Master have change with is no compatybile wih `max` level. 
Or we can ignore it in here to not awaiting for merge.

VincentLanglet:
> > > IMO we should not increase phpstan level in master yet.
> > 
> > 
> > This was increased in 3.x
> 
> I know. It should be inrease in seperate PR. Master have change with is no compatybile wih `max` level.
> Or we can ignore it in here to not awaiting for merge.

That's not an issue to fix the new static analysis error on master, like I did.
See, the test are green now.

wbloszyk (approved):

PR #469, opened 2021-03-10
DevKit updates for 2.x branch

SonataCI:


src/Block/LocaleSwitcherBlockService.php
```diff
@@ -113,7 +113,11 @@ public function configureSettings(OptionsResolver $resolver): void
 
     public function execute(BlockContextInterface $blockContext, ?Response $response = null): Response
     {
-        return $this->renderPrivateResponse($blockContext->getTemplate(), [
+        $template = $blockContext->getTemplate();
```
franmomu, line 116: does it make sense to not have a template in a block? maybe we should change `BlockContextInterface::getTemplate` to return always `string`, just asking because I don't know.

VincentLanglet: If you look about how BlockContext are created, a block can be created without a template.

But if I look at the usage, we always need a template.
There is also something wrong, because the typehint is `?string` but nothing prove that `options['template']` will be a string since we're not using an option resolver.

We could had a check `if (!is_string($template))` and throw an exception in the getter. 
https://github.com/sonata-project/SonataBlockBundle/blob/4.x/src/Block/BlockContext.php
But that would be a BC-break, isn't it ? So it require a version 5...

franmomu: yeah I didn't mean to do that in `4.x`, just pointing that out, I've opened https://github.com/sonata-project/SonataBlockBundle/issues/855.

src/Block/LocaleSwitcherBlockService.php
```diff
@@ -113,7 +113,11 @@ public function configureSettings(OptionsResolver $resolver): void
...
+
+        \assert(\is_string($template));
```
VincentLanglet, line 118: Isn't better to do
```
if (null === $template) {
    throw new \Exception()
}
```

PR #477, opened 2021-04-12
Use sonata-project/admin-bundle 4-alpha-1

franmomu:
## Changelog

```markdown
### Removed
- Remove support for `sonata-admin/block-bundle 3.x`
```


franmomu:
Not sure how we should deal with this @sonata-project/contributors: https://github.com/sonata-project/SonataTranslationBundle/pull/477/checks?check_run_id=2323311762

VincentLanglet:
Should we allow
```
"sonata-project/admin-bundle": "dev-master || 4.0.0-alpha-x",
```
for the next release ?

jordisala1991 (dismissed):

composer.json
```diff
@@ -21,22 +21,23 @@
...
+        "symfony/dependency-injection": "^4.4 || ^5.2",
+        "symfony/form": "^4.4 || ^5.2",
+        "symfony/http-foundation": "^4.4 || ^5.2",
+        "symfony/http-kernel": "^4.4 || ^5.2",
+        "symfony/options-resolver": "^4.4 || ^5.2",
         "twig/twig": "^2.12 || ^3.0"
     },
     "conflict": {
         "doctrine/annotations": "<1.8",
         "doctrine/persistence": "<1.3.4",
```
jordisala1991, line 36: There are conflicts that probably could be cleaned, or require dev with pipes too.

franmomu: I'll create another PR, we could maybe drop support for `phpcr` since https://github.com/sonata-project/SonataDoctrinePhpcrAdminBundle is abandoned.

VincentLanglet: Since we won't provide a Sonata4 support for SonataDoctrinePhpcrAdminBundle, if we want to provide a Sonata4 compatibility we had to indeed.

VincentLanglet (dismissed):

composer.json
```diff
@@ -54,9 +55,9 @@
         "psalm/plugin-phpunit": "^0.15.0",
         "psalm/plugin-symfony": "^2.0",
         "sonata-project/doctrine-orm-admin-bundle": "4.x-dev",
```
jordisala1991, line 56: can you change this one too?

franmomu: The problem with this one is if we use `^4.0`, it uses `4.0.0-alpha-1` which requires `sonata-project/admin-bundle` `4.0.0-alpha-1` and the test using `dev-master` branch doesn't work, I guess we could use that when releasing another alpha from `sonata-project/doctrine-orm-admin-bundle`.

jordisala1991: If you use minimum stability dev, it throws an error too?

franmomu: It would work, but not sure if we want to low the `minimum-stability` to `dev` to every package. 

What about the last change, I've set the stability to the affected packages instead of changing the global one.

VincentLanglet (approved):

jordisala1991 (approved):
I like the idea to only lower the minimum stability for those packages, could you apply it back to the orm?

jordisala1991:
Thank you @franmomu

PR #481, opened 2021-04-14
Raise psalm to level 2

franmomu:


psalm.xml
```diff
@@ -14,6 +14,15 @@
                 <file name="tests/Fixtures/Traits/ORM/ArticlePersonalTranslation.php"/>
             </errorLevel>
         </NonInvariantDocblockPropertyType>
+        <PropertyNotSetInConstructor>
```
VincentLanglet, line 17: Sometimes we're using the baseline, sometime we're using the issueHandler.
We should always use the same strategy isn't it ?

I think the baseline is better because there is an `occurrences` argument.

If we want to globally suppress something, issueHandler is best for.
If we want to ignore a specific issue for a specific line of code, the baseline seems better.

WDYT ?

franmomu: Well, in this case using `referencedProperty` I think it's more accurate because it's pointing to the exact property. 

With `psalm` I tend to follow https://psalm.dev/docs/running_psalm/dealing_with_code_issues/ where the baseline it's supposed to be used when there are several errors at once.

VincentLanglet: The strategy I used for SonataAdmin (https://github.com/sonata-project/SonataAdminBundle/blob/master/psalm.xml#L15-L20) was the following: 
- If we decide to voluntary ignore some type of error (for example `ContainerDependency`), it's in the issue Handler.
- If we can't fix an error, at the current moment, but it should/will be fixed some day thanks to a bugfix in psalm or an improvement in our code, it's in the baseline.

To me, the baseline is like a `TODO: FixMe`.
And it can included some PR link: https://github.com/sonata-project/SonataAdminBundle/blob/master/psalm-baseline.xml#L37

But maybe it's not the right way ? @sonata-project/contributors WDYT about where the ignored errors should be ?

VincentLanglet (approved):

jordisala1991 (approved):

PR #490, opened 2021-05-03
Set locale properly from the admin

franmomu:
## Subject

There was this code:
```php
if ($admin->hasRequest()) {
    $this->translatableLocale = (string) $admin->getRequest()->get(self::TRANSLATABLE_LOCALE_PARAMETER);
}

if (null === $this->translatableLocale) {
    $this->translatableLocale = $this->defaultTranslationLocale;
}
```

if there was request, but no parameter, `$this->translatableLocale` was set to an empty string instead of the default locale.

I am targeting this branch, because the casting was introduced in this branch.


VincentLanglet (approved):

jordisala1991 (approved):

jordisala1991:
Thank you @franmomu

PR #494, opened 2021-05-16
Ignore psalm issues

franmomu:
The issue is with the `->end()` call on a possible null value.


VincentLanglet:
Should we add a link to https://github.com/psalm/psalm-plugin-symfony/issues/174 ?

VincentLanglet (approved):

franmomu:
It is needed to fix the current build

PR #495, opened 2021-05-16
Migrate XML config files to PHP

franmomu:
The parameters in the container holding class names have been removed.

## Changelog

```markdown
### Removed
- Removed class parameters from the container 
```


VincentLanglet:
I think we did this on SonataAdmin 3.x branch.
Is there a reason to not doing this on the stable branch ?

franmomu:
> I think we did this on SonataAdmin 3.x branch.
> Is there a reason to not doing this on the stable branch ?

Way less work, in `2.x` we should keep the parameters and then when merging with `master` remove them and adjust the changes between `2.x` and `master` in these service definitions.

After this and maybe other PR I think it's ready to release an alpha version.

VincentLanglet (approved):

phansys (approved):

PR #500, opened 2021-06-05
Explicitly require symfony/intl

franmomu:
It is used in https://github.com/sonata-project/SonataTranslationBundle/blob/a7e5eb022f6b915ef878e3ae3a54b17c55e56356/src/Twig/Extension/IntlExtension.php#L16-L17

I've labeled as _pedantic_ since there is no problem using it with Symfony 4 (`symfony/form` requires it).


VincentLanglet (approved):

jordisala1991:
Thanks @franmomu

PR #502, opened 2021-07-01
DevKit updates for 2.x branch

SonataCI:


.php_cs.dist
```diff
@@ -63,6 +63,7 @@ $rules = [
 
 $finder = PhpCsFixer\Finder::create()
     ->in(__DIR__)
+    ->exclude('node_modules')
```
OskarStark, line 66: @VincentLanglet this should only be added if it has frontend, right?

VincentLanglet: I didn't add any check.

Prior to this we were excuding `Resources` or `var` even if there was no folder

PR #522, opened 2021-08-13
Make compiler passes internal and minor tweaks

franmomu:
I think with these changes we can release a RC.

~I was thinking about deprecating the abstract classes from https://github.com/sonata-project/SonataTranslationBundle/tree/master/src/Model since I don't think users should extend their models from this bundle, in any case these models should be provided by the library (this can be also deprecated in `3.x`).~ Done in https://github.com/sonata-project/SonataTranslationBundle/pull/537

I am targeting this branch, because removing method is BC break.

## Changelog

```markdown
### Changed
- Changed compiler passes as internal
- AbstractTranslatableAdminExtension is now `@internal`
```

### DIFF
--- a/src/Admin/Extension/AbstractTranslatableAdminExtension.php
+++ b/src/Admin/Extension/AbstractTranslatableAdminExtension.php
@@ -22,6 +22,8 @@ use Sonata\TranslationBundle\Model\TranslatableInterface;
  * @author Nicolas Bastien <nbastien.pro@gmail.com>
  *
  * @phpstan-extends AbstractAdminExtension<TranslatableInterface>
+ *
+ * @internal
  */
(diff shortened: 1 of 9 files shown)


VincentLanglet (approved):

VincentLanglet:
`interface Sonata\TranslationBundle\Model\TranslatableInterface` has no typehint ; is there a reason ? @franmomu 

Also public method of the abstract class must be final
- AbstractPersonalTranslatable
- AbstractTranslatable
- AbstractTranslatableAdminExtension
And we should do the same for traits, maybe (don't know if it makes senses ?)

both `TranslatableTrait::getLocale`  miss typehint too

Didn't we also do something for TwigExtension ? Like declaring the construct internal ?

VincentLanglet (approved):

PR #523, opened 2021-08-13
Remove flag related css

franmomu:
We could also add webpack to parse the css file, I'll give it a try.


franmomu:
I've removed more code since seems related to flags based on https://github.com/sonata-project/SonataTranslationBundle/pull/9 and the classes used in https://github.com/sonata-project/SonataTranslationBundle/tree/master/src/Resources/views

Without flags, it just shows a list of languages.

VincentLanglet (approved):

PR #524, opened 2021-08-13
Add strict rules

franmomu:


VincentLanglet (dismissed):
In other project we used `self::assert` but I don't think that change anything.

franmomu:
> In other project we used `self::assert` but I don't think that change anything.

Didn't see that, I've used `static` as suggested by phpstorm.

jordisala1991:
Can we use self? I would prefer to not have to think on the next bundle if I have to use self or static.

tests/Model/GedmoTest.php
```diff
@@ -29,26 +28,24 @@ public function testTranslatableModel(): void
...
+        static::assertSame('fr', $model->getLocale());
     }
 
     public function testPersonalTranslatableModel(): void
     {
         $model = new ModelPersonalTranslatable();
         $model->setLocale('fr');
 
-        $this->assertSame('fr', $model->getLocale());
-        $this->assertInstanceOf(TranslatableInterface::class, $model);
```
jordisala1991, line 42: this should stay as we done on other bundles, the issue raised by phpstans should be silenced. cc @VincentLanglet

franmomu: But what's the point of these assertions? It is a stub we've created in tests implementing that interface, IMO it's testing the `setLocale` and `getLocale` functionality provided by the abstract class (which I think we should remove by the way).

jordisala1991: In general having the assert instance of is a good idea because even if you test those methods, you could go and change the implementation interface with same methods and it wont make test fail. BUT your application could break because you rely on TranslatableInterface somehow (or maybe symfony does).

VincentLanglet (approved):

tests/Model/GedmoTest.php
VincentLanglet: Oh didn't see this. I understand both point of view.
I've created the issue on phpstan and the issue was closed with the advice to use something about BC-break.

I agree the assert instance is a good idea to be sure the same interface is always return (especially since we don't have BC-break checks). That's why I added this to the baseline with a link to the issue.
https://github.com/sonata-project/SonataAdminBundle/blob/4.x/phpstan-baseline.neon#L9
So in a general way I think those extra checks are good.

But here, indeed we're testing our test code, that's useless
https://github.com/sonata-project/SonataTranslationBundle/blob/848f70593026eea280e109661692bde4b92c5374/tests/Fixtures/Traits/ModelPersonalTranslatable.php

jordisala1991 (approved):

tests/Model/GedmoTest.php
jordisala1991: Didn't saw it was a testing code. Fine then.

jordisala1991:
Thanks @franmomu

PR #537, opened 2021-09-08
Deprecate abstract classes and traits for translatable models

franmomu:
## Subject

This bundle is supposed to handle the translation of models used in admins, but we should not provide abstract models nor traits. 

The user should not extend their models from this bundle, but instead create their own (we could also deprecate our interfaces to use the ones from the libraries).

## Changelog

```markdown
### Deprecated
- `Sonata\TranslationBundle\Model\Gedmo\AbstractPersonalTranslatable` class
- `Sonata\TranslationBundle\Model\Gedmo\AbstractPersonalTranslation` class
- `Sonata\TranslationBundle\Model\Gedmo\AbstractTranslatable` class
- `Sonata\TranslationBundle\Model\AbstractTranslatable` class
- `Sonata\TranslationBundle\Traits\Gedmo\PersonalTranslatableTrait` trait
- `Sonata\TranslationBundle\Traits\Gedmo\TranslatableTrait` trait
- `Sonata\TranslationBundle\Traits\TranslatableTrait` trait
```

### DIFF
--- a/UPGRADE-2.x.md
+++ b/UPGRADE-2.x.md
@@ -26,6 +26,19 @@ sonata_translation:
         translatable_listener_service: stof_doctrine_extensions.listener.translatable
 ```
 
+### Deprecated abstract models and traits
+
+- `Sonata\TranslationBundle\Model\Gedmo\AbstractPersonalTranslatable`.
+- `Sonata\TranslationBundle\Model\Gedmo\AbstractPersonalTranslation`.
+- `Sonata\TranslationBundle\Model\Gedmo\AbstractTranslatable`.
+- `Sonata\TranslationBundle\Model\AbstractTranslatable`.
+- `Sonata\TranslationBundle\Traits\Gedmo\PersonalTranslatableTrait`.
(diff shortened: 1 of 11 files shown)


src/Admin/Extension/AbstractTranslatableAdminExtension.php
```diff
@@ -23,6 +23,8 @@
  * @author Nicolas Bastien <nbastien.pro@gmail.com>
  *
  * @phpstan-extends AbstractAdminExtension<TranslatableInterface>
+ *
+ * @internal since 3.0
```
jordisala1991, line 27: I think this should be added on 3.0, I guess psalm with start complaining the moment you add that annotation, even though there is a comment for 3.0.

franmomu: it's already in `master`, but yeah, I've removed it just in case there is some problem.

jordisala1991 (approved):

VincentLanglet (approved):

VincentLanglet:
Thanks

PR #556, opened 2021-10-08
Add current_locale variable in LocaleSwitcherBlockService

ggabrovski:
## Subject

Change visibility of a translatable locale, because its used in the template
https://github.com/sonata-project/SonataTranslationBundle/blob/737b8d5218fbb0becb7af2bfea97cad4fe92c07a/src/Resources/views/Block/block_locale_switcher.html.twig#L11

I am targeting this branch, because BC-break.

## Changelog

```markdown
### Fixed
- Fixed accessing current locale from `block_locale_switcher.html.twig` template
```

### DIFF
--- a/src/Block/LocaleSwitcherBlockService.php
+++ b/src/Block/LocaleSwitcherBlockService.php
@@ -30,10 +30,7 @@ final class LocaleSwitcherBlockService extends AbstractBlockService
      */
     private $localeProvider;
 
-    /**
-     * NEXT_MAJOR: Make the $localeProvider mandatory
-     */
-    public function __construct(Environment $twig, ?LocaleProviderInterface $localeProvider = null)
+    public function __construct(Environment $twig, LocaleProviderInterface $localeProvider)
     {
         parent::__construct($twig);
         $this->localeProvider = $localeProvider;
@@ -48,7 +45,7 @@ final class LocaleSwitcherBlockService extends AbstractBlockService
                 'template' => '@SonataTranslation/Block/block_locale_switcher.html.twig',
                 'locale_switcher_route' => null,
                 'locale_switcher_route_parameters' => [],
-                'current_locale' => $this->localeProvider ? $this->localeProvider->get() : null,
+                'current_locale' => $this->localeProvider->get(),
             ]
         );
     }
--- a/src/Block/LocaleSwitcherBlockService.php
+++ b/src/Block/LocaleSwitcherBlockService.php
@@ -15,14 +15,27 @@ namespace Sonata\TranslationBundle\Block;
 
 use Sonata\BlockBundle\Block\BlockContextInterface;
 use Sonata\BlockBundle\Block\Service\AbstractBlockService;
+use Sonata\TranslationBundle\Provider\LocaleProviderInterface;
 use Symfony\Component\HttpFoundation\Response;
 use Symfony\Component\OptionsResolver\OptionsResolver;
+use Twig\Environment;
 
 /**
  * @author Nicolas Bastien <nbastien.pro@gmail.com>
  */
 final class LocaleSwitcherBlockService extends AbstractBlockService
 {
+    /**
+     * @var LocaleProviderInterface
+     */
+    private $localeProvider;
+
+    public function __construct(Environment $twig, LocaleProviderInterface $localeProvider)
+    {
+        parent::__construct($twig);
+        $this->localeProvider = $localeProvider;
+    }
+
     public function configureSettings(OptionsResolver $resolver): void
     {
         $resolver->setDefaults(
@@ -32,6 +45,7 @@ final class LocaleSwitcherBlockService extends AbstractBlockService
                 'template' => '@SonataTranslation/Block/block_locale_switcher.html.twig',
                 'locale_switcher_route' => null,
                 'locale_switcher_route_parameters' => [],
+                'current_locale' => $this->localeProvider->get(),
             ]
         );
     }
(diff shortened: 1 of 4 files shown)


franmomu:
Thanks @ggabrovski for reporting this, can we maybe do another thing?

Apparently, the template is to render a block, this code looks like should not belong to the template:

https://github.com/sonata-project/SonataTranslationBundle/blob/737b8d5218fbb0becb7af2bfea97cad4fe92c07a/src/Resources/views/Block/block_locale_switcher.html.twig#L10-L14

So I'd say it's better to move it to the Block service: https://github.com/sonata-project/SonataTranslationBundle/blob/3.x/src/Block/LocaleSwitcherBlockService.php

Here we can inject the `LocaleProviderInterface $localeProvider` and add in https://github.com/sonata-project/SonataTranslationBundle/blob/737b8d5218fbb0becb7af2bfea97cad4fe92c07a/src/Block/LocaleSwitcherBlockService.php#L28-L35

```php
'current_locale' => $this->localeProvider->get(),
```

ggabrovski:
Thanks @franmomu you approach is much better. Please check if it's fine like this.

src/Resources/views/Block/block_locale_switcher.html.twig
```diff
@@ -2,17 +2,11 @@
 
 {% if admin.class is translatable %}
     {% set object = block_context.settings.object %}
-    {% set currentLocale = object.locale|default(null) %}
+    {% set currentLocale = object.locale|default(block_context.settings.current_locale) %}
```
franmomu:
```suggestion
```

I would remove this line since the value was always overridden by the foreach (and `object.locale` may not exist anymore) and just replace the usage of `currentLocale` by `current_locale`.

franmomu (commented):
Looks cleaner! Just a comment and also you will need to also adjust `LocaleSwitcherBlockServiceTest.php`.

There are some unrelated tests failing that should be fixed once https://github.com/sonata-project/SonataTranslationBundle/pull/557 is merged.

ggabrovski:
Thanks @franmomu for you professional guidelines. If you find anything else, please let me know.

franmomu:
Code looks good! 

Just two things, it would be nice if you squash all your commits since they are all related to the same change and also modify the issue description, I've modified the changelog, but feel free to change it too.

ggabrovski:
I modified the issue description, but I made a mess trying to squash the commit and I don't know how to fix it. Can you merge it like this? Sorry.

franmomu:
> I modified the issue description, but I made a mess trying to squash the commit and I don't know how to fix it. Can you merge it like this? Sorry.

No problem, I've rebased, squash the commits and the history was weird, so I've just cherry-picked that commit and applied to a fresh `3.x`.

Thanks for the work done!

franmomu (dismissed):

ggabrovski:
Thank you very much!

core23 (approved):

src/Block/LocaleSwitcherBlockService.php
```diff
@@ -15,14 +15,27 @@
...
  * @author Nicolas Bastien <nbastien.pro@gmail.com>
  */
 final class LocaleSwitcherBlockService extends AbstractBlockService
 {
+    /**
+     * @var LocaleProviderInterface
+     */
+    private $localeProvider;
+
+    public function __construct(Environment $twig, LocaleProviderInterface $localeProvider)
```
VincentLanglet, line 33: Isn't this a BC-break @franmomu @core23 ?

VincentLanglet (changes requested):

src/Block/LocaleSwitcherBlockService.php
ChristianGripp: You're right. The `LocaleProviderInterface` should be nullable to stay BC.

core23 (changes requested):

src/Block/LocaleSwitcherBlockService.php
VincentLanglet: @ggabrovski You have to do
```
/**
 * NEXT_MAJOR: Make the $localeProvider mandatory
 */
public function __construct(Environment $twig, ?LocaleProviderInterface $localeProvider = null)
```

But then I assume we don't add `'current_locale'` in the `configureSettings` if the locale provider is null ?

ggabrovski:
Hi @VincentLanglet, I applied your correction, please check it. Do we need to make correction in the template to cover the case where the `current_locale` is null?

VincentLanglet (dismissed):

src/Block/LocaleSwitcherBlockService.php
franmomu: not really, this is not the stable branch, the visibility of  the `translatableLocale` of the extensions was change in this branch, in `2.x` works fine without this change.

core23 (dismissed):

VincentLanglet:
Is there a reason for using the "major" label @franmomu ?

franmomu:
> Is there a reason for using the "major" label @franmomu ?

this one: https://github.com/sonata-project/SonataTranslationBundle/pull/556#discussion_r725814672

franmomu (changes requested):
This is not the stable branch, there is no need to keep BC, it was fine as it was.

VincentLanglet:
> This is not the stable branch, there is no need to keep BC, it was fine as it was.

Oh sorry, now we renamed the master branch, it's not easy to know if the branch is the stable one or the unstable one. We should try something to do.

@ggabrovski I'm sorry, can you remove the latest commit ; my bad

ggabrovski:
Hi @VincentLanglet , 
no problem, it's done.

VincentLanglet (approved):

franmomu (approved):

franmomu:
Thanks @ggabrovski !

PR #557, opened 2021-10-09
Set locale to gedmo listener

franmomu:
## Subject

When using our gedmo translatable listener, if the locale is not set, it uses the its default one which is `en_US`.

This call is also made in the listener defined by [StofDoctrineExtensionsBundle](https://github.com/stof/StofDoctrineExtensionsBundle/blob/7322d49bd7ff408e44704c0cd939889ddbb490f6/src/Resources/config/translatable.xml#L19-L21)

I am targeting this branch, because this is a bugfix.

## Changelog

```markdown
### Fixed
- Setting the proper locale to Gedmo translatable listener
```

### DIFF
--- a/tests/App/DataFixtures/GedmoCategoryFixtures.php
+++ b/tests/App/DataFixtures/GedmoCategoryFixtures.php
@@ -15,7 +15,6 @@ namespace Sonata\TranslationBundle\Tests\App\DataFixtures;
 
 use Doctrine\Bundle\FixturesBundle\Fixture;
 use Doctrine\Persistence\ObjectManager;
-use Gedmo\Translatable\Entity\Repository\TranslationRepository;
 use Gedmo\Translatable\Entity\Translation;
 use Sonata\TranslationBundle\Tests\App\Entity\GedmoCategory;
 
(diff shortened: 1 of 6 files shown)


jordisala1991 (dismissed):

tests/App/DataFixtures/GedmoCategoryFixtures.php
```diff
@@ -15,27 +15,28 @@
...
+use Gedmo\Translatable\Entity\Repository\TranslationRepository;
+use Gedmo\Translatable\Entity\Translation;
 use Sonata\TranslationBundle\Tests\App\Entity\GedmoCategory;
 
 final class GedmoCategoryFixtures extends Fixture
 {
     public const CATEGORY = 'category_novel';
 
+    /**
+     * @psalm-suppress RedundantCondition psalm does not need the assertion of the repository class
```
VincentLanglet: Then, why the assertion is here ?

franmomu: for phpstan: https://github.com/sonata-project/SonataTranslationBundle/runs/3845413745?check_suite_focus=true

I can add `phpstan-ignore-next-line` instead.

VincentLanglet: I recommend instead the config 
```
    doctrine:
        objectManagerLoader: phpstan-object-manager-loader.php
        allowNullablePropertyForRequiredField: true
```
in the phpstan neon.

And the object manager loader will have
```
<?php

use Sonata\TranslationBundle\Tests\App\AppKernel;

require __DIR__.'/vendor/autoload.php';

$kernel = new AppKernel();
$kernel->boot();

return $kernel->getContainer()->get('doctrine')->getManager();
```

VincentLanglet: And the assert wont be necessary

VincentLanglet (approved):

PR #558, opened 2021-10-11
Drop support for PHP 7.3

franmomu:
## Changelog

```markdown
### Removed
- Dropped support for PHP 7.3
```

### DIFF
--- a/src/Admin/Extension/AbstractTranslatableAdminExtension.php
+++ b/src/Admin/Extension/AbstractTranslatableAdminExtension.php
@@ -33,20 +33,11 @@ abstract class AbstractTranslatableAdminExtension extends AbstractAdminExtension
      */
     public const TRANSLATABLE_LOCALE_PARAMETER = 'tl';
 
-    /**
-     * @var string|null
-     */
-    protected $translatableLocale;
+    protected ?string $translatableLocale = null;
 
-    /**
-     * @var TranslatableChecker
-     */
-    protected $translatableChecker;
+    protected TranslatableChecker $translatableChecker;
 
-    /**
-     * @var LocaleProviderInterface
-     */
-    private $localeProvider;
+    private LocaleProviderInterface $localeProvider;
 
     public function __construct(TranslatableChecker $translatableChecker, LocaleProviderInterface $localeProvider)
     {
--- a/tests/App/Entity/GedmoCategory.php
+++ b/tests/App/Entity/GedmoCategory.php
@@ -21,28 +21,21 @@ use Gedmo\Translatable\Translatable;
 class GedmoCategory implements Translatable
 {
     /**
+     * @Gedmo\Locale
+     */
+    public ?string $locale = null;
+    /**
      * @ORM\Id
      * @ORM\Column(type="string")
      * @ORM\GeneratedValue(strategy="NONE")
-     *
-     * @var string
      */
-    private $id;
+    private string $id;
 
     /**
      * @Gedmo\Translatable(fallback=true)
      * @ORM\Column(type="string")
-     *
-     * @var string
      */
-    private $name;
-
-    /**
-     * @Gedmo\Locale
-     *
-     * @var string|null
-     */
-    private $locale = null;
+    private string $name;
 
     public function __construct(string $id = '', string $name = '')
     {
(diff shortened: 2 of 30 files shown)


src/Admin/Extension/AbstractTranslatableAdminExtension.php
```diff
@@ -33,25 +33,17 @@ abstract class AbstractTranslatableAdminExtension extends AbstractAdminExtension
      */
     public const TRANSLATABLE_LOCALE_PARAMETER = 'tl';
 
-    /**
-     * @var string|null
-     */
-    protected $translatableLocale;
+    protected ?string $translatableLocale;
```
ChristianGripp:
```suggestion
    protected ?string $translatableLocale = null;
```

tests/App/Entity/GedmoCategory.php
```diff
@@ -75,10 +69,7 @@ public function setName(string $name): void
         $this->name = $name;
     }
 
-    /**
-     * @param string $locale
-     */
-    public function setLocale($locale): void
+    public function setLocale(string $locale): void
```
ChristianGripp: This looks strange. The property is nullable, but the setter isn't

tests/Fixtures/Model/ModelTranslatable.php
```diff
@@ -20,10 +20,8 @@ class ModelTranslatable implements TranslatableInterface
 {
     /**
      * @Gedmo\Locale()
-     *
-     * @var string|null
      */
-    private $locale;
+    private ?string $locale = null;
 
     public function setLocale(string $locale): void
```
ChristianGripp: Same here

tests/Fixtures/Traits/ORM/ArticlePersonalTranslatable.php
```diff
@@ -34,49 +35,37 @@ class ArticlePersonalTranslatable
...
     public function __construct()
     {
         $this->translations = new ArrayCollection();
     }
 
-    /**
-     * @param string $locale
-     */
-    public function setLocale($locale): void
+    public function setLocale(string $locale): void
```
ChristianGripp: And here

core23 (changes requested):

tests/App/Entity/GedmoCategory.php
franmomu: I've created https://github.com/sonata-project/SonataTranslationBundle/pull/559

src/Admin/Extension/AbstractTranslatableAdminExtension.php
franmomu: Should we set `null` in the property or in the constructor or both? 

I usually do it in the constructor except for entities which could not be instantiated through the constructor.

phpstan.neon.dist
```diff
@@ -9,9 +9,4 @@ parameters:
         - src
         - tests
     treatPhpDocTypesAsCertain: false
-    checkGenericClassInNonGenericObjectType: true
```
VincentLanglet: With @jordisala1991 we added this in every project to be more explicit about what is checked ; I would prefer to keep it personnaly

core23 (dismissed):

phpstan.neon.dist
franmomu: Those options are enabled by default when reaching level 6 and the other one is enabled when using `strict-rules`. 
IMHO there is no need to add them explicitly, it can be confusing to see them explicitly set (it did to me).

VincentLanglet: On this other side, if you don't know what is the default value, it could be helpful.
Since we've added this to every config, it can be confusing to have it on some and not the others, like the rule was set for some and not all.

cc @jordisala1991 WDYT ?

jordisala1991: We should discuss that on a separate PR.

VincentLanglet: Is it ok for you @franmomu to revert that change, at least temporary, in order to merge the drop of 7.3 ?

Maybe we should open an issue on dev kit about this. We may also consider using dev-kit to dispatch part of the config of phpstan/psalm if we have the same everywhere

VincentLanglet (dismissed):

jordisala1991:
Rebasing should fix the build @franmomu

jordisala1991 (approved):

jordisala1991:
Or not need to rebase...

VincentLanglet (approved):

VincentLanglet:
Thanks

PR #559, opened 2021-10-12
Remove TranslatableInterface::getLocale

franmomu:
## Subject

There is no need to have this method, as we do in `TranslatableAdminExtension::alterObject()` we directly set the locale when creating the object.

Ideally, I think we should remove the interface since `gedmo` do not enforce to use one, we could use the `TranslatableListener` to fetch the configuration for an entity and set the value with reflection as they do.

I am targeting this branch, because this breaks BC.

## Changelog

```markdown
### Removed
- Removed `TranslatableInterface::getLocale()` method
```

### DIFF
--- a/tests/App/Entity/GedmoCategory.php
+++ b/tests/App/Entity/GedmoCategory.php
@@ -42,7 +42,7 @@ class GedmoCategory implements TranslatableInterface
      *
      * @var string|null
      */
-    private $locale;
+    private $locale = null;
 
     public function __construct(string $id = '', string $name = '')
     {
@@ -75,10 +75,7 @@ class GedmoCategory implements TranslatableInterface
         $this->name = $name;
     }
 
-    /**
-     * @param string $locale
-     */
-    public function setLocale($locale): void
+    public function setLocale(string $locale): void
     {
         $this->locale = $locale;
     }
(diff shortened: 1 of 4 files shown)


tests/App/Entity/GedmoCategory.php
```diff
@@ -40,14 +40,15 @@ class GedmoCategory implements TranslatableInterface
     /**
      * @Gedmo\Locale
      *
-     * @var string|null
+     * @var string
```
VincentLanglet: Isnt better to use null when no value is set ?

franmomu: What do you mean by better? I personally don't like to have nullable strings when working with entities, but I don't mind changing it here if necessary.

VincentLanglet: I don't like using `''` as no value. To me `''` is a value when `null` means no value.
Elsewhere in the project, we generally use `null` to means no value.

It would be the same than initializing the int to `0`.

But I won't block it if others are ok with this.

core23 (approved):

VincentLanglet (approved):

VincentLanglet:
Thanks

PR #599, opened 2021-12-08
Remove UserLocaleSubscriber

franmomu:
## Subject

~I'll try to finish it this week, but the idea is:~

`UserLocaleSubscriber` IMO is too specific, it has nothing to do with this bundle (translations) and the user can follow https://symfony.com/index.php/doc/4.4/session/locale_sticky_session.html#setting-the-locale-based-on-the-user-s-preferences to create its own.

`LocaleSubscriber` is needed because it stores the locale in session.

I am targeting this branch, because this is BC break.

## Changelog

```markdown
### Removed
- `UserLocaleSubscriber` class in favor of implementing its own solution
```


VincentLanglet (approved):

jordisala1991 (approved):

jordisala1991:
Thanks!

PR #636, opened 2022-05-07
Translatable Checker user is_a since

mpoiriert:
The translatable checker should user is_a instead of instance of since it can receive a string ($object is not necessarily a 'object)

## Subject

I am targeting this branch, because it's a bug, $object is a string in some of the flow.

## Changelog

```markdown
### Changed
 - Sonata\TranslationBundle\Checker\TranslationChecker rely on 'is_a' instead of insteance of to fully support string parameter.

```

## To do

- [x] Update the tests;
- [ ] Update the documentation;
- [ ] Add an upgrade note.

### DIFF
--- a/src/Checker/TranslatableChecker.php
+++ b/src/Checker/TranslatableChecker.php
@@ -75,6 +75,8 @@ final class TranslatableChecker
      * Check if $object is translatable.
      *
      * @param object|string|null $object
+     *
+     * @phpstan-param object|class-string|null $object
      */
     public function isTranslatable($object): bool
     {
@@ -82,20 +84,22 @@ final class TranslatableChecker
             return false;
         }
 
-        $objectInterfaces = class_implements($object);
-        \assert(\is_array($objectInterfaces));
-        foreach ($this->getSupportedInterfaces() as $interface) {
-            if (\in_array($interface, $objectInterfaces, true)) {
-                return true;
-            }
-        }
-
-        foreach ($this->getSupportedModels() as $model) {
-            if ($object instanceof $model) {
+        foreach ($this->getSupportedClasses() as $interface) {
+            if (is_a($object, $interface, true)) {
                 return true;
             }
         }
 
         return false;
+    }
+
+    /**
+     * @return string[]
+     *
+     * @phpstan-return class-string[]
+     */
+    private function getSupportedClasses(): array
+    {
+        return array_merge($this->getSupportedInterfaces(), $this->getSupportedModels());
     }
 }
--- a/src/Twig/Extension/SonataTranslationExtension.php
+++ b/src/Twig/Extension/SonataTranslationExtension.php
@@ -59,6 +59,8 @@ final class SonataTranslationExtension extends AbstractExtension
      * Check if $object is translatable.
      *
      * @param object|string|null $object
+     *
+     * @phpstan-param object|class-string|null $object
      */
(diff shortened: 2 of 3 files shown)


franmomu:
Thanks @mpoiriert! can you please add a test using a `class-string` and it would be nice to add `class-string` in the phpdoc.

mpoiriert:
@franmomu I have made the change you've ask as well as increasing testing coverage.

I don't know exactly why the models and interfaces are split in too. I would recommend deprecating them and just use a "setClasses" or something like this. This is out of scope but that could be done in another task.

VincentLanglet:
Seems like CS and Static ANalysis is failing

src/Checker/TranslatableChecker.php
```diff
@@ -71,27 +71,31 @@ public function getSupportedModels(): array
         return $this->supportedModels;
     }
 
+    /**
+     * @return string[]
+     *
+     * @phpstan-return class-string[]
+     */
+    public function getSupportedClasses(): array
```
franmomu: I think this should be `private` if it's not used outside the scope of this class

src/Checker/TranslatableChecker.php
```diff
@@ -71,27 +71,31 @@ public function getSupportedModels(): array
...
+    {
+        return array_merge($this->getSupportedInterfaces(), $this->getSupportedModels());
+    }
+
     /**
      * Check if $object is translatable.
      *
      * @param object|string|null $object
+     *
+     * @phpstan-param class-string $object
```
franmomu:
```suggestion
     * @phpstan-param object|class-string|null $object
```

franmomu:
> I don't know exactly why the models and interfaces are split in too. I would recommend deprecating them and just use a "setClasses" or something like this. This is out of scope but that could be done in another task.

There were also traits before, I guess they were added one by one, but yeah, that can be done in another PR.

VincentLanglet (changes requested):
The .idea are removed from the commit.

@jordisala1991 should we add `.idea` to our gitignore or let the developper configure a gitignore global ? I'm not against adding it to our gitignore, it's not a big cost for us and helps contributions.

SonataTranslationBundle.iml
```diff
@@ -0,0 +1,12 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
VincentLanglet: I don't know what this is but this shouldnt be committed

mpoiriert: It's also a ide related file. I will fix it

mpoiriert:
@VincentLanglet Sorry for the .idea folder it's a mistake from my part. I didn't that I could add gitignore file from outside the folder structure. Issue is that I cannot do it globally since in the private repository of my client we commit those a per recommendation from the ide vendor. Obviously it would help every contributors if they don't know about this but it could be better to document the global git ignore (or point to a documentation) that way people learn thing at the same time :)

I have a bigger concern about linter since the xml lint fail locally because of the xml file in the .idea folder. So I cannot run the xml linter properly and using make lint will fail at xml and the ones after that will not be executed, I have to run them manually and ignore the xml.

VincentLanglet (approved):

franmomu (approved):

franmomu:
Thanks @mpoiriert!

PR #650, opened 2022-08-06
Try without requiring the knp extensions

VincentLanglet:
This should be installed by the extension installer


VincentLanglet:
Need https://github.com/KnpLabs/DoctrineBehaviors/pull/709

PR #692, opened 2023-06-01
DevKit updates for 3.x branch

SonataCI:


jordisala1991:
Maybe @phansys or @franmomu can help to solve this one...

annotation_reader is not available by default on 5.4 anymore (it seems).

How should someone define the translatable listener from gedmo?

```
        ->set('app.gedmo.translation_listener', TranslatableListener::class)
            ->call('setAnnotationReader', [service('annotation_reader')])
            ->call('setDefaultLocale', [param('locale')])
            ->call('setTranslationFallback', [false])
            ->tag('doctrine.event_subscriber');
```

I think we should use by default attributes in tests. And in src we might need to do some adjustments to, because it is using only annotations.

phansys:
I'll try to take a look on this :+1:

jordisala1991:
I merge those changes since the failure is already present without them.

Any help is welcome into fixing this.

PR #693, opened 2023-06-03
Add support for BlockBundle 5.0

jordisala1991:
## Subject

I am targeting this branch, because this is bc.

## Changelog

```markdown
### Added
- Support for SonataBlockBundle 5.0
```


jordisala1991:
Looks strange to me that this works now... @phansys @VincentLanglet to me there is still something to update here since we are relying on doctrine annotations instead of attributes for gedmo usage.

But I wont block this PR if the builds are passing.

PR #721, opened 2024-11-10
Do not require knplabs/doctrine-behaviors

franmomu:
To allow updates from dev-kit, there are lots of SA errors to ignore... not sure if it is worth it


VincentLanglet (approved):

VincentLanglet:
I never used this bundle ; but if it doesn't work with knplabs/doctrine-behaviors, is it still useful @franmomu ?

PR #733, opened 2025-02-13
Fixed error 'You have requested a non-existent parameter "locale"'

hithran:
## Subject

Fixed an issue with the use of the 'locale' parameter no longer being present in new Symfony versions.

I am targeting this branch, because I've implemented the fix while maintaining backwards compatibility

## Changelog

```markdown
### Fixed
error 'You have requested a non-existent parameter "locale"' when using Symfony 7
```


VincentLanglet:
Thanks

